PR #15, opened 2019-04-03
Initial implementation of async pool

elprans:
The majority of the code is a copy of asyncpg.pool.

### DIFF
--- a/edgedb/asyncio_pool.py
+++ b/edgedb/asyncio_pool.py
@@ -1,0 +1,729 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2019-present MagicStack Inc. and the EdgeDB authors.
+#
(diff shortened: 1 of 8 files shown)


edgedb/asyncio_pool.py
```diff
@@ -0,0 +1,722 @@
...
+    def terminate(self):
+        """Terminate all connections in the pool."""
+        if self._closed:
+            return
+        self._check_init()
+        for ch in self._holders:
+            ch.terminate()
+        self._closed = True
+
+    async def expire_connections(self):
```
1st1: Do we need this now?

elprans: Yes.  This is useful when you want to change the connection address in your pool and ensure all connections reconnect gracefully.

PR #17, opened 2019-04-05
Stop defaulting to a particular transaction isolation level

elprans:
If a transaction isolation level is not specified, don't assume any
defaults and just generate a `START TRANSACTION` statement without an
explicit `ISOLATION LEVEL`.  This allows the connection and the server
to be free in their choice of the default.

### DIFF
--- a/edgedb/transaction.py
+++ b/edgedb/transaction.py
@@ -30,7 +30,7 @@ class TransactionState(enum.Enum):
     FAILED = 4
 
 
-ISOLATION_LEVELS = {'read_committed', 'serializable', 'repeatable_read'}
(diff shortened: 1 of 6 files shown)


edgedb/transaction.py
```diff
@@ -39,22 +39,11 @@ class BaseTransaction:
                  '_state', '_nested', '_id', '_managed')
 
     def __init__(self, connection, isolation, readonly, deferrable):
```
1st1: Do we need to fix nested transactions? Ie see if the isolation level/readonly flag/deferred flag diverge from what is set for the outer block?

PR #25, opened 2019-06-21
Add type hints for connections and transactions

### LINKED ISSUE #23: Type annotations
Hi!

First, thanks for all the great work on EdgeDB. The Python driver looks fab, and I especially like the built-in async support.

---

**Problem description**
From what I've seen, `edgedb-python` public-facing APIs do not provide type annotations. I discovered this by using `edgedb.async_connect()` and wondering why my editor wouldn't autocomplete the `con` object.

**Proposed solution**
Since `edgedb-python` is targeting Python 3.6+, I believe it should fully embrace type hints. Adding these would add a lot in terms of IDE support and developer productivity.

**Implementation ideas**
If not, I'd be very glad to give a helping hand adding these to the library. The PR might be quite big — should we consider splitting the work by tackling public-facing APIs only first? Let me know!

florimondmanca:
Fixes #23 

Preview of what this looks like (VSCode):

Screenshot 2019-06-22 at 00 20 26

Screenshot 2019-06-21 at 23 41 39

Screenshot 2019-06-21 at 23 41 26

Screenshot 2019-06-21 at 23 54 59

Notes:

- I used `Set` and `Object` for the return types of `.fetchall()` and `.fetchone()`, but I'm not 100% sure they're the good ones to use.
- The built-in datatypes such as `Set`, `Object`, `Link`, etc. are defined in Cython, so obviously IDE's can't pick them up for autocomplete. I believe we could find a way to provide MyPy stub files for these(see [examples](https://github.com/python/mypy/wiki/Creating-Stubs-For-Python-Modules#examples)), but not sure whether that's something IDEs can pick up. Thankfully, the [documentation](https://edgedb.com/docs/clients/00_python/api/types) on datatypes is very thorough, but I'm still thinking it would be really nice to get autocompletion for these as well. Any thoughts?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -281,6 +281,7 @@ setuptools.setup(
     provides=['edgedb'],
     zip_safe=False,
     include_package_data=True,
+    package_data={'edgedb': ['py.typed']},
     ext_modules=[
         distutils_extension.Extension(
             "edgedb.pgproto.pgproto",
--- a/edgedb/__init__.py
+++ b/edgedb/__init__.py
@@ -25,14 +25,15 @@ from edgedb.datatypes.datatypes import Tuple, NamedTuple, EnumValue
 from edgedb.datatypes.datatypes import Set, Object, Array, Link, LinkSet
 from edgedb.datatypes.datatypes import Duration
 
-from .asyncio_con import async_connect
+from .asyncio_con import async_connect, AsyncIOConnection
 from .asyncio_pool import create_async_pool, AsyncIOPool
-from .blocking_con import connect
+from .blocking_con import connect, BlockingIOConnection
 from .transaction import Transaction, AsyncIOTransaction
 
-
 __all__ = (
-    'async_connect', 'connect', 'create_async_pool', 'AsyncIOPool',
+    'async_connect', 'AsyncIOConnection',
+    'connect', 'BlockingIOConnection',
+    'create_async_pool', 'AsyncIOPool',
     'EnumValue', 'Tuple', 'NamedTuple', 'Set',
     'Object', 'Array', 'Link', 'LinkSet',
     'Transaction', 'AsyncIOTransaction',
--- a/edgedb/asyncio_con.py
+++ b/edgedb/asyncio_con.py
@@ -19,12 +19,14 @@
 
 import asyncio
 import time
+import typing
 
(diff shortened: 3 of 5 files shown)


elprans:
> I used Set and Object for the return types of .fetchall() and .fetchone(), but I'm not 100% sure they're the good ones to use.

`Set` is correct for `fetchall()`, but `fetchone()` may return any type, not just `Object`.

> I believe we could find a way to provide MyPy stub files for these(see examples), but not sure whether that's something IDEs can pick up.

[PEP 561](https://www.python.org/dev/peps/pep-0561/) might help.

florimondmanca:
> Set is correct for fetchall(), but fetchone() may return any type, not just Object.

Got it, I pushed an update.

> PEP 561 might help.

Yup; I added `py.typed`.

Also, it turns out that Jedi (which is used in many editors I believe) released support for `.pyi` files just yesterday (https://github.com/davidhalter/jedi/issues/839#issuecomment-504175752), which is unbelievably timely.

However after upgrading, I wasn't able to get it working on VSCode straight away (e.g. by adding a `__init__.pyi` with a stub for `Set`). To be honest, stub files are still kind of over my head. I don't really understand how they work/who uses it and how, and I can't seem to find good resources online to how it all fits together.

Anyhow, I think having type annotations on these pure-Python types (connections and transactions) is a good step forward already.

setup.py
```diff
@@ -281,6 +281,7 @@ def finalize_options(self):
     provides=['edgedb'],
     zip_safe=False,
     include_package_data=True,
+    package_data={'edgedb': ['py.typed']},
```
1st1: This is great, thank you for working on this!  The only question I have is why is this line necessary and what does it do.

florimondmanca: So apparently (I mean, according to PEP 561) the py.typed file allows type checkers to know that this package contains type annotations, and use them when checking dependant packages that import this one.

This basically only has an effect when running mypy against the package, but I noticed that actually most IDEs don’t really care. I’m not sold myself on how intuitive this looks, but it’s in the spec, so 🤷‍♂️.

Interesting related discussion: https://github.com/encode/starlette/pull/455

1st1:
Merged! Thanks for the PR!

PR #26, opened 2019-07-14
Correct syntax errors in Python API code sample.

mvaled:
There was a ";" before LIMIT and an extra closing parenthesis.


1st1:
Thank you!

PR #29, opened 2019-10-17
Update usage of asyncio.get_event_loop()

aeros:
The motivation for this PR was from a recent discussion with @1st1 regarding the usage of `asyncio.get_event_loop()`. 

Within the context of coroutines, it is [recommended](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop) to use `asyncio.get_running_loop()` instead of `asyncio.get_event_loop()`. This is because the behavior or `asyncio.get_event_loop()` is significantly more complex and implicit in comparison to `asyncio.get_running_loop()`. The [performance](https://gist.github.com/aeros/b328f04f15af68543466f41dc2be118a) difference between the two is fairly negligible; with `asyncio.get_running_loop()` being slightly faster.

Additionally, I updated the examples in the docs to use `asyncio.run(coro)` instead of using `loop = asyncio.get_event_loop(); loop.run_until_complete(coro)` or `loop.get_event_loop().run_until_complete(coro)`. `asyncio.run()` was provisionally added to the asyncio API in 3.7, but it is no longer provisional.

### DIFF
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -102,11 +102,11 @@ Connection
 
         >>> import asyncio
         >>> import edgedb
-        >>> async def run():
+        >>> async def main():
         ...     con = await edgedb.async_connect(user='edgedeb')
         ...     print(await con.fetchone('SELECT 1 + 1'))
         ...
-        >>> asyncio.get_event_loop().run_until_complete(run())
+        >>> asyncio.run(main())
         {2}
 
 
--- a/edgedb/asyncio_con.py
+++ b/edgedb/asyncio_con.py
@@ -206,7 +206,7 @@ async def async_connect(dsn: str = None, *,
                         connection_class=None,
                         timeout: int = 60) -> AsyncIOConnection:
 
-    loop = asyncio.get_event_loop()
+    loop = asyncio.get_running_loop()
 
     if connection_class is None:
         connection_class = AsyncIOConnection


docs/api/asyncio_con.rst
```diff
@@ -102,11 +102,11 @@ Connection
 
         >>> import asyncio
         >>> import edgedb
-        >>> async def run():
+        >>> async def main():
```
aeros: This change is an extra, non-functional addition. IMO, it makes significantly more sense logically to use `asyncio.run()` on a coroutine named *main*, rather than using `asyncio.run()` on a coroutine named *run*.

docs/usage.rst
```diff
@@ -204,8 +204,7 @@ Below is an example of a connection pool usage:
         return app
 
 
-    loop = asyncio.get_event_loop()
-    app = loop.run_until_complete(init_app())
```
1st1: This is unfortunately incorrect. We'll need to probably update the example to use the latest aiohttp APIs. 

The old code was initializing an event loop, and running the `init_app()` coroutine in that event loop.  Then aiohttp was reusing the same loop in `web.run_app()`.

The new code (your edits) is running `init_app()` in one loop, and using yet another event loop in `web.run_app()`.

It's a bit convoluted and not obvious, but the core reason is that this example uses old aiohttp APIs.  Anyways, please revert this change.

aeros: > It's a bit convoluted and not obvious, but the core reason is that this example uses old aiohttp APIs. Anyways, please revert this change.

Done, thanks for the explanation.

1st1:
Kyle, please squash your commits into one with a clear commit message.

Also, if you want to fix the aiohttp example in the docs to use asyncio.run & modern APIs your help would be appreciated!

aeros:
> Kyle, please squash your commits into one with a clear commit message.

Okay, will do. This will be good practice, since in CPython this is usually done by the core developer that merges the changes. 

> Also, if you want to fix the aiohttp example in the docs to use asyncio.run & modern APIs your help would be appreciated!

Sounds good! I have some familiarity with aiohttp, but I'll have to spend some time reading its API documentation and do some testing to verify that the example code works properly while functioning as intended.

1st1:
> Okay, will do. This will be good practice, since in CPython this is usually done by the core developer that merges the changes.

We try to care more about good commit messaged (especially in the main edgedb repo). So we tend to ask contributors to squash commits manually (when it makes sense).

aeros:
@1st1 
> Kyle, please squash your commits into one with a clear commit message.

Done. Let me know if the commit message needs to be improved at all.

1st1:
Thank you, Kyle!

1st1:
Sorry, Kyle, we have to revert the change. `get_running_loop` isn't available in Python 3.6 (it's a new 3.7 function).  My mistake.  We have to support 3.6.

aeros:
> Sorry, Kyle, we have to revert the change. get_running_loop isn't available in Python 3.6 (it's a new 3.7 function).

@1st1 Oh, okay. I had thought it was added it 3.6. Would you like me to open a separate PR to just include the documentation changes or did you only revert the code change to `async_connect()`?

1st1:
I only reverted the get_running_loop stuff: https://github.com/edgedb/edgedb-python/commit/12e98e3a1096f6ef5f37218aeb2337fe4f8331b8

1st1:
We'll have a proper ci testing multiple Python versions in a day or two and this kind of hiccup shouldn't ever happen again. Sorry for the revert, again, my fault as a reviewer :(

aeros:
@1st1 
> We'll have a proper ci testing multiple Python versions in a day or two and this kind of hiccup shouldn't ever happen again. 

I'd be glad to help with CI debugging at some point, that's a decent amount of what I've been doing with CPython recently. Specifically, finding problematic consistent or intermittent failures in the CI and working on PRs to fix the regression tests. That's how the process of adding `asyncio.shutdown_default_executor()` started, from an intermittent CI failure. I find that maintaining stable CI goes a long way. 

> Sorry for the revert, again, my fault as a reviewer :(

No worries at all! It certainly happens from time to time, and I don't mind at all. Especially since I was still able to contribute to the documentation, even if it was in a fairly minor way. (:

PR #30, opened 2019-10-17
Add sanity check to scram.XOR

elprans:
The XOR operation must always receive inputs of equal length, so add an
explicit sanity check for that.


Neustradamus:
Good job!

Linked to:
- https://github.com/scram-sasl/info/issues/1

PR #44, opened 2019-11-18
Upgrade flake8 and fix pycodestyle version

tailhook:
Also added ignore of B006, because it warns in the following way:
```
./_testbase.py:52:72: B006 Do not use mutable data structures for argument defaults.  They are created during function definition time. All calls to the function reuse this one instance of that data structure, persisting changes between them.
./_cluster.py:126:39: B006 Do not use mutable data structures for argument defaults.  They are created during function definition time. All calls to the function reuse this one instance of that data structure, persisting changes between them.
```
Which is pretty safe at that instances.

**Or maybe fix specific instances?**

----

Without the upgrade it fails like this, because pycodestyle is not frozen (and upgraded since I've installed from scratch):
```
                                                        
    yield func(self.plugins[name], *args, **kwargs)                                                                                                           
  File "/usr/local/lib/python3.7/dist-packages/flake8/plugins/manager.py", line 421, in load_plugin                                                           
    return plugin.load_plugin()                                                                                                                               
  File "/usr/local/lib/python3.7/dist-packages/flake8/plugins/manager.py", line 186, in load_plugin                                                           
    raise failed_to_load                                                                                                                                      
flake8.exceptions.FailedToLoadPlugin: Flake8 failed to load plugin "pycodestyle.break_after_binary_operator" due to module 'pycodestyle' has no attribute 'break_after_binary_operator'.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,11 @@ CYTHON_DEPENDENCY = 'Cython==0.29.13'
 
 # Minimal dependencies required to test edgedb.
 TEST_DEPENDENCIES = [
-    'flake8~=3.5.0',
+    # pycodestyle is a dependency of flake8, but it must be frozen because
+    # their combination breaks too often
+    # (example breakage: https://gitlab.com/pycqa/flake8/issues/427)
+    'pycodestyle~=2.5.0',
+    'flake8~=3.7.9',
     'uvloop>=0.12.0rc1;platform_system!="Windows"',
 ]
 
--- a/edgedb/_cluster.py
+++ b/edgedb/_cluster.py
@@ -123,7 +123,7 @@ class Cluster:
 
         return edgedb.connect(**connect_args)
 
-    def init(self, *, server_settings={}):
+    def init(self):
         cluster_status = self.get_status()
 
         if not cluster_status.startswith('not-initialized'):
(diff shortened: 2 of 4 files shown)


.flake8
```diff
@@ -1,3 +1,3 @@
 [flake8]
-ignore = E402,E731,D100,D101,D102,D103,D104,D105,W503,W504,E252
+ignore = E402,E731,D100,D101,D102,D103,D104,D105,W503,W504,E252,B006
```
1st1: I wouldn't ignore it -- in general mutable defaults isn't the best practice. 

In this specific case those parameters aren't used anymore and should just be removed.

setup.py
```diff
@@ -43,7 +43,11 @@
 
 # Minimal dependencies required to test edgedb.
 TEST_DEPENDENCIES = [
-    'flake8~=3.5.0',
+    # pycodestyle is a dependency of flake, but it must be fixed because
+    # their combination breaks too often
+    # (example breakage: https://gitlab.com/pycqa/flake8/issues/427)
+    'pycodestyle~=2.5.0',
+    'flake8~=3.7.9',
```
1st1: 👍

1st1:
Thanks!

PR #52, opened 2019-12-01
Add support for the std::bigint scalar type

elprans:
`std::bigint` is an explicit variant of a numeric with zero fractional
scale.  It is an error to pass any value with fractional part as an
argument to a query accepting `bigint`.

I also took an opportunity to clean up and organize client-side argument
validation error reporting.  Invalid argument values now consistently
raise `InvalidArgumentValueError` which includes information on which
argument is problematic, as well as an excerpt from the value repr().


edgedb/protocol/codecs/codecs.pyx
```diff
@@ -425,6 +448,11 @@ cdef register_base_scalar_codecs():
     register_base_scalar_codec(
         'std::decimal',
         pgproto.numeric_encode_binary,
+        decimal_decode)
+
+    register_base_scalar_codec(
+        'std::bigint',
+        bigint_encode,
         pgproto.numeric_decode_binary)
```
1st1: We want bigint to be decoded as Python int.

edgedb/protocol/codecs/codecs.pyx
```diff
@@ -381,6 +379,31 @@ cdef duration_decode(pgproto.CodecContext settings, FRBuffer *buf):
...
+cdef decimal_decode(pgproto.CodecContext settings, FRBuffer *buf):
+    return pgproto.numeric_decode_binary_ex(settings, buf, True)
+
+
+cdef bigint_encode(pgproto.CodecContext settings, WriteBuffer buf, obj):
+    if isinstance(obj, float):
+        raise TypeError(
+            f'expected an integer or an integral Decimal'
+        )
+    elif isinstance(obj, decimal.Decimal):
```
1st1: I don't think we should have any association with decimal for std::bigint -- can't we just use Python's `int` type?

elprans: I'd thought about that.  One complication is that we cannot distinguish regular ints from bigints in repl.  I can make a subclass of int to work this around, would that do?

1st1: >  I can make a subclass of int to work this around, would that do?

Two ways to solve this:

* A subclass. We can try that now.

* Add more introspection utilities to edgedb-python (currently private API) to introspect the codec of any value being rendered.

What I'm certain about is that repl shouldn't guide this decision at all -- `std::bigint` should be a Python `int` as it's just a simpler type to work with for users.

tailhook: Why you need this in repl? Just for `123n` in repr? If just for that we might skip it for now. As its trivial in rust and we're going to make repl rustish in Alpha 3, I think.

elprans: OK. I've made it decode as a regular `int`.  Accepting integral instances of `decimal.Decimal` is still fine, IMO.

edgedb/__init__.py
```diff
@@ -110,6 +110,7 @@
     QueryArgumentError,
     MissingArgumentError,
     UnknownArgumentError,
+    InvalidArgumentValueError,
     NoDataError,
 )
-# </ERRORS-AUTOGEN>
+# </ERRORS-AUTOGEN>
```
1st1: Can you fix the tool to add that newline if needed?

edgedb/protocol/codecs/codecs.pyx
1st1: > OK. I've made it decode as a regular int. Accepting integral instances of decimal.Decimal is still fine, IMO.

Isn't it a lot like accepting integral floats? Isn't this a mis-feature that we add just because we can?

edgedb/protocol/codecs/codecs.pyx
```diff
@@ -381,6 +379,36 @@ cdef duration_decode(pgproto.CodecContext settings, FRBuffer *buf):
...
+        if obj.as_tuple().exponent < 0:
+            raise TypeError(
+                f'expected an integer or an integral Decimal'
+            )
+    elif not isinstance(obj, int):
+        try:
+            obj = int(obj)
```
1st1: This means we also accept strings. Compare this with the check we have for other ints:

```
        if type(obj) is not int and hasattr(type(obj), '__int__'):
```

You should do the same thing here.

edgedb/protocol/codecs/codecs.pyx
elprans: No. `Decimal` instances with `exponent >= 0` can be considered bigints.  I.e we do not accept `Decimal('1.0'), but we _do_ accept `Decimal('12345')`. Whereas there is no such thing as integral float `1.0` is still a float.

The cost of supporting this is practically zero, and I don't think it's a misfeature.  That said, if you are strongly against, I won't be dying on this hill.

1st1: > if you are strongly against, I won't be dying on this hill.

Let me put it this way: I'm not strongly against, but I don't understand why this is needed.

Especially in Python where the builtin `int` is already unlimited precision and people don't use decimals for this (even in Python 2 there was the `long` type).

elprans: Ironically, in other int encoders, the check is *weaker*. `if type(obj) is not int and hasattr(type(obj), '__int__'):` happily allows `Decimal` instances, even those with `exponent < 0`:

```python
print(con.fetchone('SELECT <int64>$arg', arg=decimal.Decimal('10.10')))
```

This returns `10`.  We obviously need to fix this, but then to prohibit `decimal` we need to *actively* ban it, which, I think is a bit pointless.

elprans: Actually, this is even more interesting. We have explicit casts from floats / decimals to ints in EdgeQL.  So question: why should the encoder be more strict than a cast?

edgedb/__init__.py
elprans: Fixed

edgedb/protocol/codecs/codecs.pyx
elprans: Fixed

edgedb/protocol/codecs/codecs.pyx
1st1: > We obviously need to fix this,

Yes

> We have explicit casts from floats / decimals to ints in EdgeQL. So question: why should the encoder be more strict than a cast?

Explicit, not implicit.  Allowing `fetchone('SELECT <int64>$1', 1.2)` would be similar to an implicit cast from float to int, something that we don't have in edgeql.

elprans: Well, technically, that `<int64>` _is_ explicit, but I see the point.

1st1: I look at casts for arguments as a hint for the argument's type, not actual casts. At least I remember us defining them that way. 

If/when in the future we have a better type analyses we should be able to infer parameter types in expressions like `1 + $a`, where `$a` would be an `int64`. If you relax the semantics now this feature will be extremely confusing, as the result of `fetchone(`SELECT 1 + $a`, a=1.2)` call would be `2`, not `2.2`. If we reject `1.2` now we won't have this problem.

elprans: OK, I've made passing a non-int an error.

PR #55, opened 2019-12-06
Fix tests using cal::* types

tailhook:


tailhook:
@elprans do we have a nightly build for macos?

elprans:
> do we have a nightly build for macos?

[Yes, we do](https://packages.edgedb.com/macos/edgedb-1-alpha2_1.0a2.dev360+g2204fdb5_2019120600.nightly.pkg).  Though it seems there's no convenient stable URL for it.

PR #56, opened 2019-12-09
Use timedelta for std::duration, rename Duration -> RelativeDelta

tailhook:
The RelativeDelta itself is not used and not documented, for now, but it should be enabled shortly so I decided to keep it in the codebase.

### DIFF
--- a/docs/api/types.rst
+++ b/docs/api/types.rst
@@ -42,7 +42,8 @@ The table below shows the correspondence between EdgeDB and Python types.
 | ``datetime``         | offset-aware :py:class:`datetime.datetime \         |
 |                      | <python:datetime.datetime>`                         |
 +----------------------+-----------------------------------------------------+
-| ``duration``         | :py:class:`edgedb.Duration`                         |
+| ``duration``         | :py:class:`datetime.timedelta \                     |
+|                      | <python:datetime.timedelta>`                        |
 +----------------------+-----------------------------------------------------+
 | ``float32``,         | :py:class:`float <python:float>`                    |
 | ``float64``          |                                                     |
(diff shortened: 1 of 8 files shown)


agritheory:
Nice! I like the Duration namespace better than RelativeDelta, but that's python's fault. To make this a non-breaking change, could you not have the existing Duration class issue a deprecation warning rather than remove it entirely?
Disclaimer: I'm just an innocent bystander and not formally associated with the project...

tailhook:
> To make this a non-breaking change, could you not have the existing Duration class issue a deprecation warning rather than remove it entirely?

Current EdgeDB doesn't have `cal::relativedelta` type (it's on our to-do list, but after alpha 2). The `std::duration` doesn't allow `months` and `days` parameters like old duration had.

So if we have an alias `Duration=RelativeDelta` you can't use (i.e. send to EdgeDB) the type anyway. And if we have an alias `Duration=timedelta` then constructor signature is different so your code needs adjustments anyway. 

So it's a breaking change anyway, and it's better to crash on import rather than getting random errors all over the place.

tailhook:
```
 + wget https://packages.edgedb.com/macos/edgedb-1-alpha2_latest.nightly.pkg -O edgedb.pkg -nv
https://packages.edgedb.com/macos/edgedb-1-alpha2_latest.nightly.pkg:
2019-12-09 17:28:10 ERROR 502: Bad Gateway.
##[error]Process completed with exit code 8.
```
@elprans, should I retry?

elprans:
@tailhook k8s shows no issues and I can download the file.  Not sure what happened there.  Try again.

vpetrovykh:
This change is breaking the edgedb server code. E.g. [here](https://github.com/edgedb/edgedb/blob/c5c7a7b32d05a2b21a947c521e46e53568f503ab/edb/repl/render_binary.py#L313), but it's probably worth looking for other places.

tailhook:
Yes, that's expected. In the hindsight, I should have notified the team better.

PR #57, opened 2019-12-12
Version v0.7.0a3

tailhook:
Breaking Changes
================

* ``std::duration`` type is now represented as ``datetime.timedelta``


ambv:
@elprans, note, pushing the tag failed here in CI.

PR #58, opened 2019-12-19
Add support for dump/restore protocols

1st1:


1st1:
This is also ready for review, @elprans

PR #60, opened 2019-12-27
Fix SCRAM implementation to adhere spec

tailhook:
By specification `nonce` is not base64 encoded string. Quoting the
spec:

> r: This attribute specifies a sequence of random printable ASCII
>     characters excluding ',' (which forms the nonce used as input to
>     the hash function).  No quoting is applied to this string.

I mean current edgedb doesn't accept anything other than valid
base64-encoded string including padding, which is wrong. (The generation
of nonce using base64 as well as client implementation are fine).

Unfortunately this requires a tiny patch in the edgedb itself (so tests are broken now).

### DIFF
--- a/edgedb/scram/__init__.py
+++ b/edgedb/scram/__init__.py
@@ -38,8 +38,8 @@ def generate_salt(length: int = DEFAULT_SALT_LENGTH) -> bytes:
     return os.urandom(length)
 
 
-def generate_nonce(length: int = RAW_NONCE_LENGTH) -> bytes:
-    return os.urandom(length)
+def generate_nonce(length: int = RAW_NONCE_LENGTH) -> str:
+    return B64(os.urandom(length))
 
 
 def build_verifier(password: str, *, salt: typing.Optional[bytes] = None,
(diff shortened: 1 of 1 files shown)


edgedb/scram/__init__.py
```diff
@@ -187,16 +187,18 @@ def parse_client_first_message(resp: bytes):
     if nonce_attr[0:1] != b'r':
         raise ValueError('malformed SCRAM message')
 
-    _, _, nonce_b64 = nonce_attr.partition(b'=')
-    nonce = base64.b64decode(nonce_b64)
+    _, _, nonce_bin = nonce_attr.partition(b'=')
+    nonce = nonce_bin.decode('ascii')
+    if not nonce.isprintable():
```
ambv: If you'd like to be thorough, reject commas here, too.

edgedb/scram/__init__.py
```diff
@@ -329,8 +332,10 @@ def parse_server_first_message(msg: bytes):
...
```
ambv: Ditto.

edgedb/scram/__init__.py
tailhook: The original string is split by commas, so commas are not possible here.

edgedb/scram/__init__.py
tailhook: Answered in the previous comment.

Neustradamus:
@tailhook: Good job!

Linked to:
- https://github.com/scram-sasl/info/issues/1

PR #64, opened 2019-12-31
Use 32-bit length prefixed strings in descriptors

tailhook:
This depends on [PR in pgproto](https://github.com/MagicStack/py-pgproto/pull/3) and [PR in edgedb](https://github.com/edgedb/edgedb/pull/1031).

### DIFF
--- a/edgedb/protocol/codecs/codecs.pyx
+++ b/edgedb/protocol/codecs/codecs.pyx
@@ -63,7 +63,7 @@ cdef class CodecsRegistry:
             bytes tid = frb_read(spec, 16)[:16]
             uint16_t els
             uint16_t i
-            uint16_t str_len
+            uint32_t str_len
             uint16_t pos
             int32_t dim_len
             BaseCodec res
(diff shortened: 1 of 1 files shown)


tailhook:
I need to release this before we can merge edgedb and tests obviously will not work before that.

edgedb/protocol/codecs/codecs.pyx
```diff
@@ -84,7 +84,7 @@ cdef class CodecsRegistry:
                 els = <uint16_t>hton.unpack_int16(frb_read(spec, 2))
                 for i in range(els):
                     frb_read(spec, 1)
-                    str_len = <uint16_t>hton.unpack_int16(frb_read(spec, 2))
+                    str_len = <uint32_t>hton.unpack_uint32(frb_read(spec, 4))
```
elprans: You don't need to cast anymore.  That was the whole reason to add `uint` unpack variants, wasn't it?

tailhook: Yes. You're right. There are so many casts in the code that they look like a convention here :)

PR #66, opened 2020-01-06
Disallow passing int-like objects as <int*> arguments

### LINKED ISSUE #53: Disallow passing objects with __int__ as <int> arguments

ambv:
Fixes #53

### DIFF
--- a/edgedb/protocol/codecs/codecs.pyx
+++ b/edgedb/protocol/codecs/codecs.pyx
@@ -377,15 +377,21 @@ cdef duration_decode(pgproto.CodecContext settings, FRBuffer *buf):
     return datetime.timedelta(microseconds=microseconds)
 
 
+cdef checked_decimal_encode(
+    pgproto.CodecContext settings, WriteBuffer buf, obj
+):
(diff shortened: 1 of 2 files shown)


agritheory:
> Please also restrict inputs to `decimal` to accept integers and and Decimal.

Clarification: You're referring to the GP scalar 'decimal' not the python class Decimal, correct? In my experience string-to-decimal conversion for the python Decimal object is pretty crucial.

elprans:
I'm referring to what's allowed to be passed as an argument to a query with a `std::decimal` parameter.  We chose to be strict and only allow passing Python `int` and `decimal.Decimal`.

agritheory:
Thank you for that. I have a related question about scalars, but it's better suited for the chat.

1st1:
Please squash into one commit before merging.

dimaqq:
OT: is this your new job, @ambv ? :)

ambv:
> OT: is this your new job, @ambv ? :)

It is. [Not so new anymore](https://twitter.com/llanga/status/1179354806537932802) actually. :-)

PR #68, opened 2020-01-08
Rename AsyncConnection.close() to .aclose()

### LINKED ISSUE #40: Rename `await con.close()` to `await con.aclose()`

1st1:
`close()` method in Python is a bit special as it makes an object
"closeable".  There are APIs that expect closeable objects to
shutdown when their `close` method is called. In the case of
`AsyncConnection`, old close() used to return a coroutine that
still needs to be awaited in order to do anything. Therefore, we
rename `close` to `aclose`, which also makes AsyncConnection symmetric
with async generators.

Closes #40.

### DIFF
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -224,7 +224,7 @@ Connection
             specified, the server-side default is used.
 
 
-    .. py:coroutinemethod:: close()
+    .. py:coroutinemethod:: aclose()
 
         Close the connection gracefully.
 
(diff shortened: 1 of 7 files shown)


tailhook:
Isn't `wait_closed()` a [common pattern](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.Server.wait_closed)?

1st1:
Yes, it's used when close() is synchronous. The pattern itself is an anti pattern though, I just need to figure out how to kill it in asyncio.

tailhook:
I mean the `wait_closed` name should be used. Or it depends on something?

*(I personally still don't get why everybody is obsessed by async close, I would like everything to be closed immediately, and yes acquired connections can probably be closed on release, but I'm not sure I need to wait for that when calling pool.close)*

elprans:
> I mean the wait_closed name should be used. Or it depends on something?

`await wait_closed()` is too many `waits` in my opinion.

PR #79, opened 2020-02-03
Json elements mode

tailhook:
Here is some intermediary code I need comments for.

Questions are:
1. I cant define `cdef enum IoFormat` that is visible from python, can I?
2. Is this approach of using python enum fine?
3. Do I understand correctly, that the whole `protocol` and so `SansIOProtocol` is considered private? So I can change arguments of the `execute_anonymous` method freely? (There is no dash in the import path, but `_protocol` property looks like private?

(If (3) is true, should we hide some of that modules or classes behind underscore?)


elprans:
> 1. I cant define `cdef enum IoFormat` that is visible from python, can I?

You can if you declare it as [`cpdef`](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#c-variable-and-type-definitions)

> 2. Is this approach of using python enum fine?

Yes.

> 3. Do I understand correctly, that the whole protocol and so SansIOProtocol is considered private?

Yes.

> So I can change arguments of the execute_anonymous method freely?

Yes.

tailhook:
> You can if you declare it as cpdef

Still couldn't import (unless this is a problem of something that wasn't rebuilt). Will investigate again tomorrow.

edgedb/protocol/__init__.py
```diff
@@ -15,3 +15,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import enum
+
+
+class IoFormat(enum.Enum):
```
1st1: You can use a Cython enum and put it in protocol.pxd

tailhook:
Okay, I've fixed my build process. This should be ready now.

edgedb/asyncio_con.py
```diff
@@ -28,6 +28,7 @@
 
 from .datatypes import datatypes
 from .protocol import asyncio_proto
+from .protocol.protocol import IoFormat
```
1st1: Everything looks good, one nit: can you replace `from .protocol.protocol import IoFormat` with `from .protocol import protocol` and use `IoFormat` as `protocol.IoFormat.BINARY`?

The reason is that we always import modules in all our Python projects.

tailhook: Okay. Fixed.

1st1:
Don't forget to squash :)

PR #86, opened 2020-02-06
Fix protocol upgrade handling

tailhook:
The details are:
1. Ensure that major protocol version is the same
2. Ensure minimum minor protocol version (possible lower than current)
3. Store real minor version in protocol object for further reference

Not sure how to make test for this. But I've tested on 0.8 locally, and on 0.7 is tested on github (which is considered an older protocol since this revision)


elprans:
> Not sure how to make test for this.

We should add a command line switch/config option to the server to be able to specify the protocol version explicitly.

edgedb/protocol/protocol.pxd
```diff
@@ -94,6 +94,7 @@ cdef class SansIOProtocol:
 
         readonly bytes last_status
         readonly bytes last_details
+        readonly uint16_t minor_protocol_version
```
elprans: it would be more convenient to work with the protocol version as a tuple, i.e. `(major, minor)`.

tailhook: Then it will be an `object` in cython, right? Which is subpar I think. We can make a property that returns tuple if we find some use case for that.

tailhook:
> We should add a command line switch/config option to the server to be able to specify the protocol version explicitly.

Do you mean protocol should be downgraded by server, or do you mean a flag to send arbitrary protocol number?

edgedb/protocol/protocol.pxd
elprans: It will be a `tuple`.  If you are worried about performance in a hot spot, we can always extract the components into integer components there.  The point is that doing version comparison checks as `self.version < (major, minor)` is more convenient and less error-prone.

elprans:
> Do you mean protocol should be downgraded by server

I mean that an option can be used to specify the maximum (and maybe minimum) version of the protocol that the server will *advertise*.  This will allow testing the behavior of clients against older server versions.

elprans:
> Do you mean protocol should be downgraded by server

I mean that an option can be used to specify the maximum (and maybe minimum) version of the protocol that the server will *advertise*.  This will allow testing the behavior of clients against older server versions.

edgedb/protocol/protocol.pxd
tailhook: Okay, fixed.

tailhook:
> > Do you mean protocol should be downgraded by server
> 
> I mean that an option can be used to specify the maximum (and maybe minimum) version of the protocol that the server will _advertise_. This will allow testing the behavior of clients against older server versions.

As far as I understand it's not for this PR anyway, because we have chicken and the egg problem (server still supports a single protocol until this PR is applied and integrated into EdgeDB).

PR #88, opened 2020-02-14
Fix version check

tailhook:
This commit treats `x` in `0.x` version as a major version

PR #89, opened 2020-02-18
Fix argument tuples in the protocol

tailhook:
This PR requires https://github.com/edgedb/edgedb/pull/1235 to be applied, fixes edgedb/edgedb#1220 and includes #88 because that is a prerequisite.


edgedb/protocol/protocol.pyx
```diff
@@ -106,7 +106,7 @@ cdef class SansIOProtocol:
 
         self.server_settings = {}
         self.reset_status()
-        self.protocol_version = (PROTO_VER_MAJOR, PROTO_VER_MINOR_MIN)
+        self.protocol_version = (PROTO_VER_MAJOR, 0)
```
elprans: Are you setting `minor` to `0` to force negotiation?

tailhook: No. Basically, because I don't know what to set. It should be never read before negotiation happened. Not sure setting it to None is a good idea...

tailhook:
ping! This has to be merged to unbreak current master of edgedb-python.

edgedb/protocol/protocol.pyx
```diff
@@ -807,7 +807,8 @@ cdef class SansIOProtocol:
                 self.parse_headers()
                 self.buffer.finish_message()
 
-                if major != PROTO_VER_MAJOR or minor < PROTO_VER_MINOR_MIN:
+                if (major != PROTO_VER_MAJOR or
+                        (major == 0 and minor != PROTO_VER_MINOR)):
```
fantix, line 811: `major == 0` is always `True`, because `PROTO_VER_MAJOR` is `0`? I'm asking because I'm bumping `PROTO_VER_MAJOR` to `1`, and I don't want to miss a check here.

tailhook: The idea is that when "major" is zero minor version is treated as major.

So, yes, drop the second check when you bump `PROTO_VER_MAJOR` to >= 1.

PR #91, opened 2020-03-05
Fix the test setup script.

vpetrovykh:
Make the get_setup_script match the same method from edb.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
-ignore = E402,E731,D100,D101,D102,D103,D104,D105,W503,W504,E252
+ignore = B008,B306,E203,E402,E731,D100,D101,D102,D103,D104,D105,W503,W504,E252,F999,F541
 exclude = .git,__pycache__,build,dist,.eggs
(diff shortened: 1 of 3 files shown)


vpetrovykh:
This looks like an oversight. 

Generally I wanted to use the test framework from edgedb in the tutorial, so that I can set up and tear down a test database and all that. So I was rather happy to find all the familiar base test classes in `edgedb`. I assume that we actually want these test classes duplicated between `edgedb` and `edb`.

vpetrovykh:
> So what does this change "fix"? The new code seems to be almost 100% equivalent to the old code.

The migration command was all wrong using old-style migrations: not wrapping things in modules and incorporating module name into migration name.

1st1:
> The migration command was all wrong using old-style migrations: not wrapping things in modules and incorporating module name into migration name.

The commit message doesn't reflect that... Please amend and merge.

vpetrovykh:
I've updated the `.flake8` with the same ignore codes as we use in main `edgedb` repo just to be consistent and to shut up complaints about f-strings without actual interpolated args. Aside from that I also updated the code to wrap migrations into transactions (which was giving me problems when run with `make quicktest`).

1st1:
Thanks, Victor!

PR #93, opened 2020-04-20
add information about missed keyword arguments

### LINKED ISSUE #83: Better error for mismatched query parameters and arguments
The code like:
```
await c.fetchall("""
    INSERT Foo {
        smth := <str>$smth,
        other := <datetime>$other,
        ...
        # a dozen lines skipped
        ...
        quux := <int16>$quux
    }
""", **some_dict)
```

Fails with:

```
Traceback (most recent call last):
  File "/Users/dima.tisnek/edge/ingest.py", line 47, in <module>
    asyncio.run(main())
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/Users/dima.tisnek/edge/ingest.py", line 14, in main
    await c.fetchall("""
  File "/Users/dima.tisnek/Library/Caches/pypoetry/virtualenvs/edge-tfNuXqpZ-py3.9/lib/python3.9/site-packages/edgedb/asyncio_con.py", line 99, in fetchall
    return await self._protocol.execute_anonymous(
  File "edgedb/protocol/protocol.pyx", line 534, in execute_anonymous
  File "edgedb/protocol/protocol.pyx", line 279, in _execute
  File "edgedb/protocol/protocol.pyx", line 987, in edgedb.protocol.protocol.SansIOProtocol.encode_args
  File "edgedb/protocol/./codecs/./namedtuple.pyx", line 74, in edgedb.protocol.protocol.NamedTupleCodec.encode_kwargs
RuntimeError: expected 18 keyword arguments, got 17
```

Clearly the keys in `somedict` don't match the query parameters...

It would be awesome to know which keys :)

nsidnev:
Fix #83

### DIFF
--- a/edgedb/protocol/codecs/namedtuple.pyx
+++ b/edgedb/protocol/codecs/namedtuple.pyx
@@ -71,14 +71,15 @@ cdef class NamedTupleCodec(BaseNamedRecordCodec):
 
         objlen = len(obj)
         if objlen != len(self.fields_codecs):
-            raise RuntimeError(
-                f'expected {len(self.fields_codecs)} keyword arguments, '
-                f'got {objlen}')
+            raise self._make_missing_args_error_message(obj)
 
         elem_data = WriteBuffer.new()
         for i in range(objlen):
             name = datatypes.record_desc_pointer_name(self.descriptor, i)
-            arg = obj[name]
+            try:
+                arg = obj[name]
+            except KeyError:
+                raise self._make_missing_args_error_message(obj) from None
 
             elem_data.write_int32(0)  # reserved bytes
             if arg is None:
@@ -114,3 +115,27 @@ cdef class NamedTupleCodec(BaseNamedRecordCodec):
         codec.fields_codecs = fields_codecs
 
         return codec
+
+    def _make_missing_args_error_message(self, args):
+        required_args = set()
+
+        for i in range(len(self.fields_codecs)):
+            name = datatypes.record_desc_pointer_name(self.descriptor, i)
+            required_args.add(name)
+
+        passed_args = set(args.keys())
+        missed_args = required_args - passed_args
+        extra_args = passed_args - required_args
+
+        error_message = f'expected {required_args} keyword arguments'
+        error_message += f', got {passed_args}'
+
+        missed_args = set(required_args) - set(passed_args)
+        if missed_args:
+            error_message += f', missed {missed_args}'
+
+        extra_args = set(passed_args) - set(required_args)
+        if extra_args:
+            error_message += f', extra {extra_args}'
+
+        return errors.QueryArgumentError(error_message)
--- a/tests/test_async_fetch.py
+++ b/tests/test_async_fetch.py
@@ -488,6 +488,51 @@ class TestAsyncFetch(tb.AsyncQueryTestCase):
                 'select <datetime>$0;',
                 date)
 
+    async def test_async_mismatched_args_01(self):
+        with self.assertRaisesRegex(
+                edgedb.QueryArgumentError,
+                r"expected {'a'} keyword arguments, got {'[bc]', '[bc]'}, "
+                r"missed {'a'}, extra {'[bc]', '[bc]'}"):
+
+            await self.con.fetchall("""SELECT <int64>$a;""", b=1, c=2)
+
+    async def test_async_mismatched_args_02(self):
+        with self.assertRaisesRegex(
+                edgedb.QueryArgumentError,
+                r"expected {'[ab]', '[ab]'} keyword arguments, "
+                r"got {'[acd]', '[acd]', '[acd]'}, "
+                r"missed {'b'}, extra {'[cd]', '[cd]'}"):
+
+            await self.con.fetchall("""
+                SELECT <int64>$a + <int64>$b;
+            """, a=1, c=2, d=3)
+
+    async def test_async_mismatched_args_03(self):
+        with self.assertRaisesRegex(
+                edgedb.QueryArgumentError,
+                "expected {'a'} keyword arguments, got {'b'}, "
+                "missed {'a'}, extra {'b'}"):
+
+            await self.con.fetchall("""SELECT <int64>$a;""", b=1)
+
+    async def test_async_mismatched_args_04(self):
+        with self.assertRaisesRegex(
+                edgedb.QueryArgumentError,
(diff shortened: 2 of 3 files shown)


edgedb/protocol/codecs/namedtuple.pyx
```diff
@@ -70,15 +70,26 @@ cdef class NamedTupleCodec(BaseNamedRecordCodec):
         self._check_encoder()
 
         objlen = len(obj)
-        if objlen != len(self.fields_codecs):
+        fields_len = len(self.fields_codecs)
+
+        passed_arguments = set(obj.keys())
```
1st1, line 75: This check is rather expensive so we only need to do it when we're certain that there's an error. So condition this with `if objlen <> fields_len:`

1st1:
Overall looks good. In addition to the other comment, please add a couple functional unittest with `assertRaisesRegex`.

edgedb/protocol/codecs/namedtuple.pyx
nsidnev: I understand that, but I thought it would be nice to raise a `RuntimeError` with description here instead of a `KeyError` that might occur at the encoding in case of such query, for example:
```python3
conn.fetchall("""SELECT <int64>$arg_name""", wrong_name=42)
```
for this reason I replaced the original check. But restored it back.

1st1: > I understand that, but I thought it would be nice to raise a RuntimeError with description here instead of a KeyError that might occur at the encoding in case of such query, for example:

I'm confused... even with my suggestion your latest PR code will raise a nice `RuntimeError`, right?

nsidnev: Well, it will raise an error if there are either many or few arguments. But if the number of passed arguments is equal to what is required in the query, then `KeyError` will be raised there.

So these queries will raise a `RuntimeError`:

*  
  ```python3
  conn.fetchall("""SELECT <int64>$arg1""", arg1=1, arg2=2)  # extra {'arg2'}
  ```

* 
  ```python3
  conn.fetchall("""SELECT <int64>$arg1""", arg2=1, arg3=2)  # missed {'arg1'}, extra {'arg2', 'arg3'}
  ```

And this one will still raise a `KeyError`:

* 
  ```python3
  conn.fetchall("""SELECT <int64>$arg1""", arg2=2)  # KeyError: 'arg1'
  ```

And I'm not sure that such a difference would be good from the point of view of the user.

Maybe I should move the code that raises the error to a separate function (`raise_missmatched_args_error(fields_codecs, passed_args)` for example)? It will be called:
1. when the comparaision of the number of passed and required arguments fails.
2. when the value is encoded in case of `KeyError`.

edgedb/protocol/codecs/namedtuple.pyx
```diff
@@ -70,15 +70,26 @@ cdef class NamedTupleCodec(BaseNamedRecordCodec):
...
-                f'got {objlen}')
+                f'expected {required_arguments} keyword arguments, '
+                f'got {passed_arguments}, missed {missed_arguments}, '
+                f'extra {extra_arguments}')
 
         elem_data = WriteBuffer.new()
-        for i in range(objlen):
-            name = datatypes.record_desc_pointer_name(self.descriptor, i)
-            arg = obj[name]
+        for i, required_arg in enumerate(required_arguments):
```
1st1, line 91: You can't iterate over `required_arguments` here. Sets in Python are unordered and we must preserve the sequential order of arguments here. So please keep the old

```
for i in range(objlen):
    name = datatypes.record_desc_pointer_name(self.descriptor, i)
```

loop here.  What I suggest to do to actually print a nice error message in all cases is to refactor the code as follows:

```
for i in range(objlen):
    name = datatypes.record_desc_pointer_name(self.descriptor, i)
    try:
        arg = obj[name]
    except KeyError:
        raise self._make_missing_arg_error_message(obj) from None
```

and then you can implement the actual validation logic in the new `_make_missing_arg_error_message` method. That method can be rather complex as it will only be called when some argument is actually missing.

You should also keep the original `if objlen != len(self.fields_codecs):`, just change it to

```
if objlen != len(self.fields_codecs):
    raise raise self._make_missing_arg_error_message(obj)
```

edgedb/protocol/codecs/namedtuple.pyx
1st1: Got it. See my new comment that suggests how we can raise a nice error in all cases.

edgedb/protocol/codecs/namedtuple.pyx
nsidnev: Thanks for your suggestions!

Moved the creation of the error and added new tests.

edgedb/protocol/codecs/namedtuple.pyx
```diff
@@ -114,3 +115,27 @@ cdef class NamedTupleCodec(BaseNamedRecordCodec):
...
+
+        missed_args = set(required_args) - set(passed_args)
+        if missed_args:
+            error_message += f', missed {missed_args}'
+
+        extra_args = set(passed_args) - set(required_args)
+        if extra_args:
+            error_message += f', extra {extra_args}'
+
+        return RuntimeError(error_message)
```
1st1, line 141: Let's change the error type from `RuntimeError` to `edgedb.QueryArgumentError`.

edgedb/protocol/codecs/namedtuple.pyx
1st1: Nice, please take a look at the one last nit comment.

1st1:
@nsidnev Nik, can you also squash all your commits into one and push it into this PR? For some reason github ci is struggling with pulling from your PR branch.

edgedb/protocol/codecs/namedtuple.pyx
nsidnev: Changed. Also squashed commits.

1st1:
And lastly, could you please rebase against the latest master branch?

nsidnev:
CI finally passed! Thanks! :tada:

1st1:
Nik, this is a great PR, thanks so much!

PR #95, opened 2020-04-30
change error on access to properties throught brackets

### LINKED ISSUE #50: Confusing KeyError on resulting row
On code like this:
```python
x = edb.fetchone("SELECT ...")
x["id"]
```
I get error like this:
```
KeyError: 'id'
```
But I think it should be something like this:
```
TypeError: 'User' object is not subscriptable
```
Because basically I should write `x.id` instead (and it works).

Or is there any legitimate case when both attribute and item access work on the same row?

(lib version `edgedb==0.7.0a2.dev0+e147670`)

nsidnev:
closes #50

### DIFF
--- a/tests/datatypes/test_datatypes.py
+++ b/tests/datatypes/test_datatypes.py
@@ -406,7 +406,7 @@ class TestObject(unittest.TestCase):
         with self.assertRaises(KeyError):
             o[0]
 
-        with self.assertRaises(KeyError):
+        with self.assertRaises(TypeError):
             o['id']
 
         self.assertEqual(set(dir(o)), {'id', 'c'})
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_datatypes.py
```diff
@@ -406,7 +406,7 @@ def test_object_1(self):
         with self.assertRaises(KeyError):
             o[0]
 
-        with self.assertRaises(KeyError):
+        with self.assertRaises(TypeError):
```
1st1, line 409: Let's use `assertRaisesRegex`

1st1: Actually never mind, I think the new tests cover this pretty clearly.

1st1:
Thank you, Nik!

PR #105, opened 2020-07-22
Allow passing `implicit_typenames` header; fix query codec caching

1st1:
* Allow passing a special 'parse' header to include the special __tname__
  computable;

* Fix the query cache to account for expected cardinality, implicit
  limit, and for implicit type names.


edgedb/protocol/protocol.pyx
```diff
@@ -80,14 +80,22 @@ cdef class QueryCodecsCache:
     def __init__(self, *, cache_size=1000):
         self.queries = LRUMapping(maxsize=cache_size)
 
-    cdef get(self, str query, IoFormat io_format):
-        return self.queries.get((query, io_format), None)
+    cdef get(
+        self, str query, IoFormat io_format,
```
elprans, line 84: fwiw, I prefer the `name: type` style for consistency with Python source.

1st1: Somehow it didn't work for me here (I really tried). But I'll be refactoring this code anyways.

elprans (approved):

PR #106, opened 2020-07-28
Use updated name for `create_async_pool()` in docs

ambv:


ambv:
Sure. Updated the docstrings. The code in tests is unrelated, it's got a local method called `create_pool` which is referenced in tests. Now it's actually better for it to be called that since it's not an async function so now it's clearly differentiated from `create_async_pool`.

1st1:
> since it's not an async function

Which is a bit of an anti-pattern :)

PR #112, opened 2020-08-14
Add support of credentials file

tailhook:
This also deprecates `admin=True` parameter and `edgedbadmin` schema, as
they aren't needed if proper permissions are set up. And was introduced
mainly for command-line tools.


1st1:
I'll take another look once https://github.com/edgedb/edgedb-rust/issues/45 is settled.

PR #113, opened 2020-08-21
Add support of connecting to instance by a name

tailhook:
This changes just the meaning of `dsn` positional argument. This is a part of what's discussed in https://github.com/edgedb/edgedb/discussions/1683 other things will follow in the next release.

This also deprecates `admin=True` parameter and `edgedbadmin` schema, as
they aren't needed if proper permissions are set up. And were introduced
mainly for command-line tools.

Supersedes #112

### DIFF
--- a/edgedb/con_utils.py
+++ b/edgedb/con_utils.py
@@ -22,8 +22,11 @@ import os
 import platform
 import typing
 import urllib.parse
+import warnings
+import pathlib
 
 from . import errors
+from . import credentials
 
 
 EDGEDB_PORT = 5656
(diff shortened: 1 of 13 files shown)


edgedb/con_utils.py
```diff
@@ -104,11 +107,24 @@ def _parse_hostlist(hostlist, port):
...
-
-    if dsn:
+    if admin:
+        warnings.warn(
+            'The "admin=True" parameter is deprecated and is scheduled to be '
+            'removed. Admin socket should never be used in applications. '
+            'Use command-line tool `edgedb` to setup proper credentials.',
+            DeprecationWarning, 4)
+
+    if dsn and dsn.startswith(("edgedb://", "edgedbadmin://")):
```
1st1, line 117: Hm, this is first time I see `edgedbadmin://` -- have we agreed to add it just yet?

1st1: Ah, I see, it's being deprecated. nvm

PR #118, opened 2020-08-30
Readme - Prefix local_date query with cal module

chrizzFTD:
Hi,

This is a very minor fix to the `local_date` type reference on the example queries from the README file.
Currently it fails with the following error:

```python
edgedb.errors.InvalidReferenceError: type 'default::local_date' does not exist
```
I was testing with `edbedb-python 0.10.0` and Alpha 5

Thanks for this project!


1st1:
Good catch! Thank you!

PR #125, opened 2020-10-26
Add Executor abstract base class

tailhook:


edgedb/abstract.py
```diff
@@ -0,0 +1,63 @@
...
+        ...
+
+    # TODO(tailhook) add *args, **kwargs, when they are supported
+    @abc.abstractmethod
+    def execute(self, query: str) -> None:
+        ...
+
+
+class Executor(abc.ABC):
+    pass
```
1st1, line 35: What is this class for? Can you add a docstring?

tailhook: Fixed inheritance and a docstring.

elprans (approved):

PR #126, opened 2020-11-03
Add github action to deploy site on changes to documentation

jaclarke:


1st1:
Hm, perhaps we should review when we want to push the updated docs. Currently, for edgedb.com we only publish the docs for the latest released version. Docs from the master branch might be about features that aren't yet available to the users who don't use nightlies. I think we'll end up implementing the version switcher and we'll have basically two versions in it: latest 1.0 release and "dev", with the former being default. We still need to figure out where to put version switchers for the drivers docs though.

elprans:
@1st1, @jaclarke any update here?

PR #132, opened 2020-12-01
Changes needed for edgedb/edgedb#2005

tailhook:
This includes:
1. Ignore extra headers instead of throwing away
2. Added `allow_capabilities` private field for requests
3. Added `_fetchall_with_headers` private method for tests

Note: we don't parse server-side headers (i.e. keep them a dict to
bytes) for performance reasons and for future extensibility. We don't
expose them to end user anyway

This includes #131 as tests don't work without it on edgedb-master


edgedb/asyncio_con.py
```diff
@@ -71,6 +71,31 @@ async def _fetchall(
...
+    async def _fetchall_with_headers(
+        self,
+        query: str,
+        *args,
+        __limit__: int=0,
+        __typeids__: bool=False,
+        __typenames__: bool=False,
+        __allow_capabilities__: typing.Optional[int]=None,
         **kwargs,
     ) -> datatypes.Set:
```
1st1, line 100: Change the type annotation to `typing.Tuple[datatypes.Set, ???]`, replacing ??? with the correct type of returned headers

edgedb/con_utils.py
```diff
@@ -261,7 +260,7 @@ def _parse_connect_dsn_and_args(*, dsn, host, port, user,
     if user is None:
         user = os.getenv('EDGEDB_USER')
         if not user:
-            user = getpass.getuser()
+            user = 'edgedb'
```
1st1, line 263: This was covered in another PR, right?

tests/test_async_fetch.py
```diff
@@ -732,35 +730,39 @@ def __int__(self):
...
-                await con2.aclose()
-
-        finally:
-            self.assertEqual(
-                await self.con.query(
-                    'select sys::advisory_unlock(<int64>$0)', lock_key),
-                [True])
+    # Advisory locks don't work/ for now
+    # Can't xfail because:
+    #     database "asyncfetch" is being accessed by other users
```
1st1, line 735: Not sure I follow the "database "asyncfetch" is being accessed by other users" -- just curious why you can't xfail.

tests/test_async_fetch.py
```diff
@@ -732,35 +730,39 @@ def __int__(self):
...
+    #         'select sys::advisory_lock(<int64>$0)',
+    #         lock_key)
+    #
+    #     try:
+    #         async with tg.TaskGroup() as g:
+    #
+    #             async def exec_to_fail():
+    #                 with self.assertRaises(ConnectionAbortedError):
+    #                     await con2.query(
+    #                         'select sys::advisory_lock(<int64>$0)', lock_key)
```
1st1, line 754: Yeah, we test it that way because using sleep() is unreliable. Making them testmode-only is OK.

tests/test_async_fetch.py
tailhook: Probably connection objects is kept somewhere in traceback object while we are trying to drop database in `finally`.

tests/test_async_fetch.py
tailhook: The only reliable way to use them is in transaction, since connection pool is implemented.

Anyway, if we move them to testmode, does this mean that we can also move sleep to testmode? As you said that the main motivation of keeping sleep is to use them in tests of the bindings.

edgedb/asyncio_con.py
tailhook: Fixed.

edgedb/con_utils.py
tailhook: Yes, I think I've mentioned that in the original PR message. Now rebased.

tailhook:
Fixed advisory lock test with the assumption that functions can be used in test mode in transaction.

1st1 (approved):

1st1 (approved):

PR #134, opened 2020-12-06
Unbreak blocking_con

elprans:
Changes in #132 modified the internal API, but `blocking_con` wasn't
updated.  Rectify that.

### DIFF
--- a/edgedb/blocking_con.py
+++ b/edgedb/blocking_con.py
@@ -97,7 +97,7 @@ class BlockingIOConnection(base_con.BaseConnection):
         )
 
     def query(self, query: str, *args, **kwargs) -> datatypes.Set:
-        return self._protocol.sync_execute_anonymous(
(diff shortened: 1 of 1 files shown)


1st1 (approved):

tailhook (changes requested):
It's done [here](https://github.com/edgedb/edgedb-python/blob/ecd15f1faf56f05a224a75edfefb4526b814e7a8/edgedb/protocol/blocking_proto.pyx#L91-L95) as we don't expose extra APIs into sync bindings. I think we have sync tests so they would have shown this issue.

Anything that actually breaks for you?

elprans:
> Anything that actually breaks for you?

Yes. Without this change tests break and all those methods return a tuple.

elprans:
Oh, I think this was a case of Cython not recompiling the *.c file.  The `Makefile` should probably just `rm *.c`, like we do in the server.

PR #135, opened 2020-12-06
Fix handling of asynchronous messages in auth subprotocol

elprans:
Asynchronous messages, such as `LogMessage` can appear at any point
during the connection, including during authentication, so make sure
that is handled properly.


1st1 (approved):

1st1:
Can you check if edgedb-js has the same problem?

PR #139, opened 2020-12-14
Switch to setup-edgedb and touchup the release workflow

elprans:


edgedb/_cluster.py
```diff
@@ -95,10 +95,7 @@ def __init__(
 
     @staticmethod
     def get_edgedb_server():
-        try:
-            return f'edgedb-server-{os.environ["EDGEDB_CI_SLOT"]}'
-        except KeyError:
-            return 'edgedb-server'
+        return 'edgedb-server'
```
tailhook, line 98: What's changed here? We still need binary with a suffix.

.github/workflows/tests.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
         os: [ubuntu-latest, macos-latest]
```
tailhook, line 19: Can we also add the following?
```
edgedb-version: [stable, nightly]
```

edgedb/_cluster.py
elprans: No, `edgedb-setup` adds the directory with unsuffixed `edgedb-server` to `PATH`.

.github/workflows/tests.yml
elprans: Good idea.

1st1 (approved):

tailhook:
Why this wasn't merged yet? I want this merged to update edgedb tests run with. Can we merge this, while marking #160 as expected failure.

elprans:
I forgot this wasn't merged.

PR #143, opened 2020-12-23
Implement wait_until_available and reconnects (part of RFC1004)

tailhook:
This is should be equivalent to edgedb/edgedb-js#82, and requires slightly updated edgedb/edgedb#2061 to include tags in `errors.txt`.

To do:
* [x] synchronous implementation


edgedb/errors/tags.py
```diff
@@ -0,0 +1,11 @@
...
+    'Tag',
+    'SHOULD_RECONNECT',
+]
+
+
+class Tag(object):
+    pass
+
+
+SHOULD_RECONNECT = Tag()
```
1st1, line 11: Let's do better than this:

```

class Tag(object):
    def __repr__(self):
        return f'<edgedb.{self.name}>'

SHOULD_RECONNECT = Tag('SHOULD_RECONNECT')
```

edgedb/errors/_base.py
```diff
@@ -79,11 +79,15 @@ class EdgeDBErrorMeta(Meta):
 class EdgeDBError(Exception, metaclass=EdgeDBErrorMeta):
 
     _code = None
+    tags = frozenset(())
```
1st1, line 82:
```suggestion
    tags = frozenset()
```

edgedb/errors/__init__.py
```diff
@@ -362,6 +381,24 @@ class ClientConnectionError(ClientError):
...
 
+class ConnectionFailedError(ClientConnectionError):
+    _code = 0x_FF_01_01_00
+
+
+class ConnectionFailedTemporarilyError(ConnectionFailedError):
+    _code = 0x_FF_01_01_01
+
+
+    tags = frozenset({SHOULD_RECONNECT})
```
1st1, line 392: There's one extra line before the `tags = ` line.

edgedb/base_con.py
```diff
@@ -101,6 +102,7 @@ def add_log_listener(
             **connection**: a Connection the callback is registered with;
             **message**: the `edgedb.EdgeDBMessage` message.
         """
+        # TODO(tailhook)
```
1st1, line 105: Can you clarify this and the following TODOs?

edgedb/asyncio_con.py
```diff
@@ -32,41 +35,160 @@
...
+            else:
+                tr, pr = await loop.create_connection(factory, *addr)
+        except socket.gaierror as e:
+            # All name resolution errors are considered temporary
+            raise errors.ConnectionFailedTemporarilyError(str(e)) from e
+        except OSError as e:
+            # TODO(tailhook) figure out error
+            message = str(e)
+            if e.errno is None:
+                errnos = extract_errno(message)
```
1st1, line 133: Why not `e.errno`?

edgedb/asyncio_con.py
```diff
@@ -32,41 +35,160 @@
...
+
+    def __init__(self, codecs_registry, query_cache):
+        self._codecs_registry = codecs_registry
+        self._query_cache = query_cache
+
+        # TODO(tailhook) temporary, this is not needed when we implement
+        # transaction part of RFC1004
+        self._top_xact = None
 
-    def __instancecheck__(cls, instance):
```
1st1, line 44: Why was this deleted?

1st1: Because it's now implemented via `abstract.AsyncIOExecutor`, right?

tailhook: No. Because `PoolConnection` is a subclass of `AsyncIOConnection` anyway. And we use two hooks (most notably `_reconnect`) to make it working for pools instead of using Proxy pattern and metaclasses.

edgedb/errors/tags.py
tailhook: Fixed.

edgedb/base_con.py
tailhook: No need for `is_closed` check. Removed the check and TODO.

edgedb/asyncio_con.py
tailhook: Because it's absent in "Multiple exceptions" case. I've added a docstring in `_extract_errno` text

tailhook:
Note: we don't put method name like `AsyncIOConnection.fetch_one` in interface errors. But:
1. This is quite rare error: using connection after `with acquire()` block ended
2. This is now visible in traceback (previous implementation raised error in wrapper in metaclass, so exact method called wasn't visible in a traceback).

edgedb/asyncio_con.py
```diff
@@ -32,41 +35,160 @@
...
+
+        while True:
+            for addr in addrs:
+                try:
+                    await asyncio.wait_for(
+                        self._connect_addr(loop, addr, params),
+                        config.connect_timeout,
+                    )
+                except asyncio.TimeoutError as e:
+                    if iteration == 1 or time.monotonic() < max_time:
```
1st1, line 94: Re `iteration == 1 or ...`: so if `wait_until_available` is 0 we'd still give it two tries?

tailhook: Iterations start from `1`. When `wait_until_available` is zero we give a one try for each of the addresses specified.

1st1 (approved):

PR #146, opened 2021-01-13
RFC1004: `retry` method implementation

tailhook:
This is the most basic implementation, i.e. it might not handle some cases of network errors. But it provides the needed API to handle that.

### DIFF
--- a/edgedb/_testbase.py
+++ b/edgedb/_testbase.py
@@ -222,7 +222,7 @@ class TestCase(unittest.TestCase, metaclass=TestCaseMeta):
         @functools.wraps(func)
         def cleanup():
             res = func(*args, **kwargs)
-            if asyncio.isfuture(res) or asyncio.iscoroutine(res):
+            if inspect.isawaitable(res):
                 self.loop.run_until_complete(res)
         super().addCleanup(cleanup)
 
--- a/edgedb/_testbase.py
+++ b/edgedb/_testbase.py
@@ -281,7 +281,7 @@ class DatabaseTestCase(ClusterTestCase, ConnectedTestCaseMixin):
                     'CONFIGURE SESSION SET __internal_testmode := true;'))
 
         if self.ISOLATED_METHODS:
-            self.xact = self.con.transaction()
+            self.xact = self.con.try_transaction()
             self.loop.run_until_complete(self.xact.start())
 
         if self.SETUP_METHOD:
@@ -338,9 +338,9 @@ class DatabaseTestCase(ClusterTestCase, ConnectedTestCaseMixin):
             if script:
                 # The setup is expected to contain a CREATE MIGRATION,
                 # which needs to be wrapped in a transaction.
-                tx = cls.con.transaction()
+                tx = cls.con.try_transaction()
                 cls.loop.run_until_complete(tx.start())
-                cls.loop.run_until_complete(cls.con.execute(script))
+                cls.loop.run_until_complete(tx.execute(script))
                 cls.loop.run_until_complete(tx.commit())
                 del tx
 
--- a/edgedb/_testbase.py
+++ b/edgedb/_testbase.py
@@ -218,6 +218,14 @@ class TestCase(unittest.TestCase, metaclass=TestCaseMeta):
                                 f'{expected_val!r})') from e
                 raise
 
+    def addCleanup(self, func, *args, **kwargs):
+        @functools.wraps(func)
+        def cleanup():
+            res = func(*args, **kwargs)
+            if asyncio.isfuture(res) or asyncio.iscoroutine(res):
+                self.loop.run_until_complete(res)
+        super().addCleanup(cleanup)
+
 
 class ClusterTestCase(TestCase):
 
--- a/edgedb/asyncio_con.py
+++ b/edgedb/asyncio_con.py
@@ -80,9 +80,13 @@ class _AsyncIOConnectionImpl:
             protocol and protocol.connected
         )
 
-    async def connect(self, loop, addrs, config, params):
+    async def connect(self, loop, addrs, config, params, *,
+                      single_attempt=False):
         addr = None
-        max_time = time.monotonic() + config.wait_until_available
+        if single_attempt:
+            max_time = 0
+        else:
+            max_time = time.monotonic() + config.wait_until_available
         iteration = 1
 
         while True:
@@ -182,16 +186,17 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
         for cb in self._log_listeners:
             self._loop.call_soon(cb, self._ensure_proxied(), msg)
 
-    async def ensure_connected(self):
+    async def ensure_connected(self, *, single_attempt=False):
         if not self._impl or self._impl.is_closed():
-            await self._reconnect()
+            await self._reconnect(single_attempt=single_attempt)
 
     # overriden by connection pool
-    async def _reconnect(self):
+    async def _reconnect(self, single_attempt=False):
         assert not self._borrowed_for, self._borrowed_for
         self._impl = _AsyncIOConnectionImpl()
         await self._impl.connect(self._loop, self._addrs,
-                                 self._config, self._params)
+                                 self._config, self._params,
+                                 single_attempt=single_attempt)
 
     async def _fetchall(
         self,
--- a/edgedb/asyncio_con.py
+++ b/edgedb/asyncio_con.py
@@ -30,7 +30,8 @@ from . import abstract
 from . import base_con
 from . import con_utils
 from . import errors
-from . import transaction
+from . import transaction as _transaction
+from . import legacy_transaction
 
 from .datatypes import datatypes
 from .protocol import asyncio_proto
@@ -164,6 +165,7 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
                          query_cache=query_cache)
         self._loop = loop
         self._impl = None
+        self._borrow = None
 
     def __repr__(self):
         if self.is_closed():
@@ -185,6 +187,7 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
 
     # overriden by connection pool
     async def _reconnect(self):
+        assert not self._borrow, self._borrow
         self._impl = _AsyncIOConnectionImpl()
         await self._impl.connect(self._loop, self._addrs,
                                  self._config, self._params)
@@ -199,6 +202,8 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
         __allow_capabilities__: typing.Optional[int]=None,
         **kwargs,
     ) -> datatypes.Set:
+        if self._borrow:
+            raise base_con.borrow_error(self._borrow)
         if not self._impl or self._impl.is_closed():
             await self._reconnect()
         result, _ = await self._impl._protocol.execute_anonymous(
@@ -225,6 +230,8 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
         __allow_capabilities__: typing.Optional[int]=None,
         **kwargs,
     ) -> typing.Tuple[datatypes.Set, typing.Dict[int, bytes]]:
+        if self._borrow:
+            raise base_con.borrow_error(self._borrow)
         if not self._impl or self._impl.is_closed():
             await self._reconnect()
         return await self._impl._protocol.execute_anonymous(
@@ -247,6 +254,8 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
         __limit__: int=0,
         **kwargs,
     ) -> datatypes.Set:
+        if self._borrow:
+            raise base_con.borrow_error(self._borrow)
         if not self._impl or self._impl.is_closed():
             await self._reconnect()
         result, _ = await self._impl._protocol.execute_anonymous(
@@ -262,6 +271,8 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
         return result
 
     async def query(self, query: str, *args, **kwargs) -> datatypes.Set:
+        if self._borrow:
+            raise base_con.borrow_error(self._borrow)
         if not self._impl or self._impl.is_closed():
             await self._reconnect()
         result, _ = await self._impl._protocol.execute_anonymous(
@@ -275,6 +286,8 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
         return result
 
     async def query_one(self, query: str, *args, **kwargs) -> typing.Any:
+        if self._borrow:
+            raise base_con.borrow_error(self._borrow)
         if not self._impl or self._impl.is_closed():
             await self._reconnect()
         result, _ = await self._impl._protocol.execute_anonymous(
@@ -289,6 +302,8 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
         return result
 
     async def query_json(self, query: str, *args, **kwargs) -> str:
+        if self._borrow:
+            raise base_con.borrow_error(self._borrow)
         if not self._impl or self._impl.is_closed():
             await self._reconnect()
         result, _ = await self._impl._protocol.execute_anonymous(
@@ -303,6 +318,8 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
 
     async def _fetchall_json_elements(
             self, query: str, *args, **kwargs) -> typing.List[str]:
+        if self._borrow:
+            raise base_con.borrow_error(self._borrow)
         if not self._impl or self._impl.is_closed():
             await self._reconnect()
         result, _ = await self._impl._protocol.execute_anonymous(
@@ -316,6 +333,8 @@ class AsyncIOConnection(base_con.BaseConnection, abstract.AsyncIOExecutor):
         return result
 
     async def query_one_json(self, query: str, *args, **kwargs) -> str:
+        if self._borrow:
+            raise base_con.borrow_error(self._borrow)
         if not self._impl or self._impl.is_closed():
             await self._reconnect()
         result, _ = await self._impl._protocol.execute_anonymous(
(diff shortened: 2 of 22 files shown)


edgedb/base_con.py
```diff
@@ -30,6 +30,17 @@
 BaseConnection_T = typing.TypeVar('BaseConnection_T', bound='BaseConnection')
 
 
+BORROW_ERRORS = {
+    'transaction':
+        "Connection object is borrowed for the transaction. "
```
1st1, line 35:
```suggestion
        "Connection object is borrowed for a transaction. "
```

edgedb/asyncio_con.py
```diff
@@ -199,6 +202,8 @@ async def _fetchall(
         __allow_capabilities__: typing.Optional[int]=None,
         **kwargs,
     ) -> datatypes.Set:
+        if self._borrow:
```
1st1, line 206: I'd rename to `self._borrowed_for`.

1st1: Also I'd make it an enum, not a string.

edgedb/retry.py
```diff
@@ -0,0 +1,87 @@
...
+def default_backoff(attempt):
+    return (2 ** attempt) * 0.1 + random.randrange(100) * 0.1
+
+
+class AsyncIOIteration(_transaction.AsyncIOTransaction):
+    def __init__(self, retry, owner):
+        super().__init__(owner)
+        self.__retry = retry
+
+    async def start(self):
```
1st1, line 21: Does this have to be a "public" method?

edgedb/retry.py
```diff
@@ -0,0 +1,87 @@
...
+        # `wait_until_available` timeout
+        await super().start()
+
+    async def __aexit__(self, extype, ex, tb):
+        try:
+            await super().__aexit__(extype, ex, tb)
+        except errors.ClientConnectionClosedError as err:
+            if ex is None:
+                # TODO(tailhook) should we figure out if err is *caused by*
+                # CancelledError and propagate the latter instead or retrying?
```
1st1, line 37: Hm, *can* it be caused by CancelledError? CancelledError is a BaseException, it shouldn't be wrapped into ClientConnectionClosedError by us.

edgedb/retry.py
```diff
@@ -0,0 +1,87 @@
...
+
+                # If we were going to commit, retry
+                if self.__retry._retry(err):
+                    return True
+                else:
+                    # if retries are exhausted we need to propagate this error
+                    raise err
```
1st1, line 47: IIRC you need to just `return False` here. The exception will be propagated by the interpreter. If you do it with a `raise` you'd likely skew its traceback.

edgedb/retry.py
```diff
@@ -85,3 +86,77 @@ def _retry(self, exc):
...
```
1st1, line 156: ditto

1st1 (commented):
almost lgtm, will take another look tomorrow.

edgedb/retry.py
tailhook: It's not the exception from body of the with. It's the exception we caught `from super().__exit__()`, it will not be reraised on return False.

edgedb/retry.py
tailhook: Here is how it happens:
```
Traceback (most recent call last):
  File "/work/tests/test_async_retry.py", line 96, in transaction1
    res = await tx.query_one('''
  File "/work/edgedb/transaction.py", line 238, in query_one
    result, _ = await self._connection_impl._protocol.execute_anonymous(
  File "edgedb/protocol/protocol.pyx", line 622, in execute_anonymous
    ret, attrs = await self._execute(in_dc, out_dc, args, kwargs)
  File "edgedb/protocol/protocol.pyx", line 362, in _execute
    await self.wait_for_message()
  File "edgedb/protocol/asyncio_proto.pyx", line 70, in wait_for_message
    raise
  File "edgedb/protocol/asyncio_proto.pyx", line 59, in edgedb.protocol.asyncio_proto.AsyncIOProtocol.wait_for_message
    await self.msg_waiter
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/work/tests/test_async_retry.py", line 108, in transaction1
    await barrier.ready()
  File "/work/edgedb/retry.py", line 34, in __aexit__
    await super().__aexit__(extype, ex, tb)
  File "/work/edgedb/transaction.py", line 155, in __aexit__
    await self.__rollback()
  File "/work/edgedb/transaction.py", line 199, in __rollback
    await self._connection_impl.execute(query)
  File "/work/edgedb/asyncio_con.py", line 153, in execute
    await self._protocol.simple_query(query)
  File "edgedb/protocol/protocol.pyx", line 533, in simple_query
    self.ensure_connected()
  File "edgedb/protocol/protocol.pyx", line 329, in edgedb.protocol.protocol.SansIOProtocol.ensure_connected
    raise errors.ClientConnectionClosedError(
edgedb.errors.ClientConnectionClosedError: the connection has been closed because an operation was cancelled on it
```
Basically it's because `__exit__` is called on `BaseExceptions` too.

But generally this code is non-controversial, regardless of the `CancelledError`:
1. I've changed it to capture any error

tailhook: On the second thought, this is where `EarlyNetworkError` needed to kick in. I.e. this is `COMMIT` branch if it failed, we have to be very careful when retrying transaction. So I need a separate tag here.

edgedb/retry.py
tailhook: This overrides the method on `Transaction`, so yes, it at least has to cover public method on the superclass.

edgedb/retry.py
tailhook: Another round: It looks like serialization error happens on the actual query rather than commit. Since this is the most important thing we want to fix I'm simplifying the code (by raising all errors during commit and swallowing all errors during rollback), and keep it as is. We can improve robustness here later.

edgedb/asyncio_con.py
tailhook: Done.

edgedb/retry.py
tailhook: Same as above.

edgedb/retry.py
1st1: Maybe we rename `Transaction` to `BaseTransaction`, make all methods private on it, and then we can add a new `Transaction` class with public proxies for those methods. Then this class can extend `BaseTransaction` and wouldn't need to have a `public start` method, which is useless.

tailhook: I think it's nice that object that user gets is a `Transaction` object, not something special. We can document it. User can do `isinstance(db, Transaction)` and it works the same in `retry` and `try_transaction`, etc.

We could also emit a different object in `__aenter__` for managed transactions which doesn't have `start()` method, but we chose not to dive too deep. I think this is very similar case.

1st1: > We could also emit a different object in __aenter__ for managed transactions which doesn't have start() method, but we chose not to dive too deep.

What I designed that API I don't think I put too much thought into what should be returned from `__aenter__`, tbh. So the decision wasn't super deliberate.

> I think it's nice that object that user gets is a Transaction object, not something special. We can document it. User can do isinstance(db, Transaction) and it works the same in retry and try_transaction, etc.

Maybe it would indeed be nice, but what's the point of calling `start()` manually inside `async with`?  Currently that seems to be allowed?

edgedb/_testbase.py
```diff
@@ -218,6 +218,14 @@ def assertRaisesRegex(self, exception, regex, msg=None,
                                 f'{expected_val!r})') from e
                 raise
 
+    def addCleanup(self, func, *args, **kwargs):
+        @functools.wraps(func)
+        def cleanup():
+            res = func(*args, **kwargs)
+            if asyncio.isfuture(res) or asyncio.iscoroutine(res):
```
1st1, line 225:
```suggestion
            if inspect.isawaitable(res):
```

edgedb/legacy_transaction.py
```diff
@@ -0,0 +1,324 @@
+#
```
1st1, line 1: Please add a comment clarifying why we have it -- essentially it's only needed to the legacy `con.transaction()` that's scheduled for removal soon.

edgedb/retry.py
```diff
@@ -0,0 +1,163 @@
...
+        self.__retry = retry
+
+    async def start(self):
+        if not self._managed:
+            raise errors.InterfaceError(
+                "Only managed retriable transactions are supported. "
+                "Use `async with transaction:`"
+            )
+        # TODO(tailhook) if this is not the first iteration suppress
+        # `wait_until_available` timeout
```
1st1, line 28: You mean if the connection broke on the second iteration we don't want to wait on connect?

1st1: If so let's implement before merging this?

edgedb/retry.py
1st1: Yeah, got it

edgedb/retry.py
```diff
@@ -0,0 +1,163 @@
+import asyncio
+import random
+import time
+
+from . import errors
+from . import transaction as _transaction
+
+
+DEFAULT_MAX_ITERATIONS = 3
```
1st1, line 9: We need to make this a configurable option for connect() etc.

tests/test_async_retry.py
```diff
@@ -0,0 +1,121 @@
...
+        async def transaction1(con):
+            async for tx in con.retry():
+                nonlocal iterations
+                iterations += 1
+                async with tx:
+                    # This magic query makes the test more reliable for some
+                    # reason. I guess this is because starting a transaction
+                    # in EdgeDB (and/or Postgres) is accomplished somewhat
+                    # lazily, i.e. only start transaction on the first query
+                    # rather than on the `START TRANSACTION`.
```
1st1, line 89: I don't think this is the reason, the snapshot should be captured as soon as you start the transaction. So 🤷‍♂️

1st1 (commented):
Almost there. I'd address one of the TODOs before merging and make number of iterations configurable.

edgedb/retry.py
tailhook: No. You either do:
```
async with tx:
  something
```
Or you do:
```
transaction.start()
transaction.commit()
```
In the former case `_managed` is set to `True`. In the latter case it keeps `False`. So for the transaction created by `retry`, latter case throws `InterfaceError` on unmanaged transaction `start()`.

edgedb/retry.py
tailhook: Well. We discussed this with @elprans and he suggested making it non-configuable for now.

The issue is:
1. RFC1004 provisions `with_retry_options` for this.
2. We don't want to add more options to `connect()` because we want to make a builder/fluent interface for that
3. Even if we want to add a temporary option to `connect()`, retry options is more complex than that, you're supposed to configure different retry options for different conditions (i.e. on deadlock and network error) and also retry interval (function).
4. Even if we want to make builder/fluent now or eventually, it makes little sense to add retry count to that initial connection options, because overriding rules, are not clear. For example, whether `conn_builder.retry_options(DEADLOCK, 2)` overrides `pool.with_retry_options(ALL, 3)`.

So as I've already said: RFC1004 provides a nice coherent way to configure transaction options. If you want to remove that part of it, it should be better specified how to configure all of that.

1st1: > RFC1004 provisions with_retry_options for this.

FWIW I realized a few weeks ago that the only argument I had against that idea doesn't hold when we have a server side pool of connections. And we do now. So I withdraw my objections to the proposed `with_*` design.

> Well. We discussed this with @elprans and he suggested making it non-configuable for now.

Yeah, everything you said sounds reasonable. Let's merge as is and we'll later make it configurable via with_ or some other well-thought API.

edgedb/retry.py
1st1: > ... So for the transaction created by retry, latter case throws InterfaceError on unmanaged transaction start().

So 

```python
async for tx in con.retry():
  async with tx:
    await tx.start()
```

Will throw an InterfaceError? That's what I want. I don't think I've seen a test for that -- if there isn't one can you add it please?

tests/test_async_retry.py
tailhook: Yes. Also what actually happens, is that second query is blocked rather than throwing a transaction error.

Also it looks like this also doesn't make the test 100% reliable :( (https://github.com/edgedb/edgedb-python/pull/146/checks?check_run_id=1709638747#step:7:1023) @elprans do you have any idea how to make this test reliable?

edgedb/retry.py
tailhook: Done.

edgedb/retry.py
tailhook: In your case it's "transaction is already started". But I've added few tests.

tailhook:
It's quite interesting that test consistently fails on python3.8 on macos only. Even though the failure looks like a race condition.

1st1:
> It's quite interesting that test consistently fails on python3.8 on macos only. Even though the failure looks like a race condition.

Are you 100% sure it's a race condition? If not can you debug? Sometimes what I think is a race condition turns out to be another quite serious bug (I doubt that's the case here though).

tests/test_async_retry.py
1st1: > is that second query is blocked

Hm, that's weird. The current server doesn't have any deliberate functionality to block, and neither does Postgres (AFAIK).

edgedb/retry.py
1st1: Great, thanks

edgedb/retry.py
```diff
@@ -0,0 +1,163 @@
...
+            raise errors.InterfaceError(
+                "Only managed retriable transactions are supported. "
+                "Use `async with transaction:`"
+            )
+        await self._start(single_connect=self.__iteration != 0)
+
+    async def __aexit__(self, extype, ex, tb):
+        try:
+            await super().__aexit__(extype, ex, tb)
+        except errors.EdgeDBError as err:
```
1st1, line 33: This method is quite different from `Iteration.__exit__`. Is that on purpose? I think they're supposed to look very much alike, no?

For example, in `Iteration.__exit__` you except a `errors.ClientConnectionClosedError`, and here you except on `errors.EdgeDBError`.

The latter logic is also different, here you just raise `err` if `ex is None`; there you have a bunch of code that checks tags and returns `True` or raises.

edgedb/retry.py
```diff
@@ -0,0 +1,163 @@
...
+
+                # If we were going to commit, retry
+                if (
+                    err.has_tag(errors.SHOULD_RETRY) and
+                    self.__retry._retry(err)
+                ):
+                    return True
+                else:
+                    # if retries are exhausted we need to propagate this error
+                    raise err
```
1st1, line 151: We need to make sure that we never mask CancelledError. It should be propagated as is.

tests/test_async_retry.py
```diff
@@ -0,0 +1,134 @@
...
+        with self.assertRaisesRegex(edgedb.InterfaceError,
+                                    r'.*the transaction is already started'):
+            async for tx in self.con.retry():
+                async with tx:
+                    await tx.start()
+
+        with self.assertRaisesRegex(edgedb.InterfaceError,
+                                    r'.*Use `async with transaction:`'):
+            async for tx in self.con.retry():
+                await tx.start()
```
1st1, line 134: Can you add a test that checks what happens if an unrelated exception is propagated out of the body of with?

e.g.:

```python
async for tx in self.con.retry():
  async with tx:
    1/0
```

tailhook: I'm not sure what should be tested, though. Exception propagation? The successful rollback? No retries?

1st1: > Exception propagation?

Yes. A very basic test that the iteration aborts and that the exception is propagated unmodified.

> The successful rollback? 

Yes, that too.

edgedb/retry.py
tailhook: Nice catch. I've simplified the logic in async case and missed sync case.

edgedb/retry.py
tailhook: We don't mask anything here. We reraise same error and only catch EdgeDB errors anyway. Also see comment above: this code is now simplified similarly to the async case.

tests/test_async_retry.py
tailhook: I think blocking is implied by `SELECT xxx FOR UPDATE`. So I assume it *might* apply to `ON DUPLICATE KEY` too.

edgedb/retry.py
```diff
@@ -0,0 +1,156 @@
...
+        self._owner = owner
+        self._iteration = 0
+        self._done = False
+        self._backoff = default_backoff
+        self._max_iterations = DEFAULT_MAX_ITERATIONS
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
```
1st1, line 69: Can you add a comment here like

```
# Note: when changing this code consider also
# updating Retry.__next__.
```

and do that for all other mirrored code? That would help someone like me later.

edgedb/retry.py
```diff
@@ -0,0 +1,156 @@
...
+        assert self._iteration + 1 < self._max_iterations, \
+            f"Extra retry {self._iteration}/{self._max_iterations}"
+        if self._iteration > 0:
+            await asyncio.sleep(self._backoff(self._iteration))
+        self._done = True
+        iteration = AsyncIOIteration(self, self._owner, self._iteration)
+        self._iteration += 1
+        return iteration
+
+    def _retry(self, exc):
```
1st1, line 81: Can we share this method implementation with `Retry._retry`? I'm kinda against DRY-fanaticism but in this case it's super hard to track differences or similarities between sync and async.

tests/test_async_retry.py
1st1: Just wanted to add that tests like this, while trivial, helped me a bunch of times in the past. Just adding `return True` somewhere in `__exit__` can break everything and it might not be easy to spot that in review. So let's have some very basic tests to ensure the correct behavior.

1st1 (approved):
I've left a few comments -- please address them and then merge when you thing the time is right. Other than that this is ready, I think. Great work!

cc @elprans you might also want to take a look at this (even if it's already merged by the time you do)

tests/test_async_retry.py
tailhook: Done

edgedb/retry.py
tailhook: You you mean put this in every method of the Transaction?

edgedb/retry.py
tailhook: Done

tailhook:
> > It's quite interesting that test consistently fails on python3.8 on macos only. Even though the failure looks like a race condition.
> 
> Are you 100% sure it's a race condition? If not can you debug? Sometimes what I think is a race condition turns out to be another quite serious bug (I doubt that's the case here though).

You're intuition is quite right. The issue was that retry interval was calculated incorrectly and was quite large. So it kinda race condition, but has nothing to do with postgres :)

This doesn't solve that `SELECT 1` issue, though (but I was wrong that it holds the query, as currently written the two transactions finish successfully without a conflict).

What does blocking is if I put a second barrier after `INSERT/UPDATE` query but before `COMMIT`. I.e. the second transaction blocks until first one commits (or rollbacks) and only then throws `TransactionSerializationError`.

All in all, it looks like tests work quite reliably with `SELECT 1`.

tests/test_async_retry.py
tailhook: See below.

PR #150, opened 2021-01-25
Add documentation on retry

tailhook:
Only async part is there. I'll make sync part when this is accepted, so that I don't need to fix same thing in two places.

### DIFF
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -658,8 +658,8 @@ See also:
 
     Represents a transaction or savepoint block.
 
-    Transactions are created by calling the
-    :py:meth:`AsyncIOConnection.transaction()` method.
+    Instances of this type are created by calling the
+    :py:meth:`AsyncIOConnection.try_transaction()` method.
 
 
     .. py:coroutinemethod:: start()
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -257,11 +257,16 @@ Connection
 
     .. py:method:: try_transaction()
 
-        Create an individual :py:class:`AsyncIOTransaction` object.
+        Execute a non-retryable transaction.
+        
+        Contrary to ``retry()``, ``try_transaction()`` will not attempt
+        to re-run the nested code block in case a retryable error happens.
+        
+        This is a low-level API and it is advised to use the ``retry()`` 
+        method instead.
 
-        Transaction may fail so it's often better to use :py:meth:`retry`
-        instead. This method is targeted for users who want to implement \
-        more specialized transaction retry loop.
+        A call to ``try_transaction()`` returns 
+        :py:class:`AsyncIOTransaction`.
 
         Example:
 
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -383,7 +383,7 @@ Connection Pools
                 async with tx:
                     await tx.query('SELECT {1, 2, 3}')
 
-    To hold on a specific connection object, use the ``pool.acquire()`` API:
+    To hold on to a specific connection object, use the ``pool.acquire()`` API:
 
     .. code-block:: python
 
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -236,9 +236,16 @@ Connection
 
     .. py:method:: retry()
 
-        Create a :py:class:`AsyncIORetry` object. This is a way to make
-        reliable transactions. See
-        :ref:`edgedb-python-asyncio-api-transaction` for more info.
+        Open a retriable transaction loop.
+       
+        This is the preferred method of initiating and running a database
+        transaction in a robust fashion.  The `retry()` transaction loop will
+        attempt to re-execute the transaction loop body if a transient error
+        occurs, such as a network error or a transaction serialization error.
+       
+        Returns an instance of :py:class:`AsyncIORetry`.
+       
+        See :ref:`edgedb-python-asyncio-api-transaction` for more details.
 
         Example:
 
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -686,7 +686,7 @@ See also:
 
     .. describe:: async with c:
 
-        start and commit/rollback the transaction or savepoint block
+        Start and commit/rollback the transaction or savepoint block
         automatically when entering and exiting the code inside the
         context manager block.
 
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -664,7 +664,7 @@ See also:
 
     .. py:coroutinemethod:: start()
 
-        Enter the transaction or savepoint block.
+        Start a transaction or create a savepoint.
 
     .. py:coroutinemethod:: commit()
 
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -624,9 +624,12 @@ Here execution of the http request be retried too. The core of the issue
 is that whenever transaction is interrupted user might have changed the
 email, so we have to redo all the work done.
 
-Generally it's recommended not to do potentially slow or long running
-external tasks within the transaction and split it into multiple
-ones.
+Generally it's recommended to not execute any long running
+code within the transaction unless absolutely necessary.
+
+Transactions allocate expensive server resources and having
+too many concurrently running long-running transactions will
+negatively impact the performance of the DB server.
 
 See also:
 * RFC1004_
@@ -721,4 +724,3 @@ See also:
         be repeated.
 
 .. _RFC1004: https://github.com/edgedb/rfcs/blob/master/text/1004-transactions-api.rst
-
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -259,8 +259,8 @@ Connection
                         value=value,
                     )
 
-        Note executing queries on ``tx`` object rather than the original
-        connection.
+        Note that we are executing queries on the ``tx`` object rather
+        than on the original pool itself.
 
     .. py:method:: try_transaction()
 
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -376,7 +376,7 @@ Connection Pools
                 async with tx:
                     await tx.query('SELECT {1, 2, 3}')
 
-    If you need to lock on specific connection, use ``async with`` block:
+    To hold on a specific connection object, use the ``pool.acquire()`` API:
 
     .. code-block:: python
 
--- a/docs/api/asyncio_con.rst
+++ b/docs/api/asyncio_con.rst
@@ -234,8 +234,51 @@ Connection
             If the results of *query* are desired, :py:meth:`query` or
             :py:meth:`query_one` should be used instead.
 
+    .. py:method:: retry()
+
+        Create a :py:class:`AsyncIORetry` object. This is a way to make
+        reliable transactions. See
+        :ref:`edgedb-python-asyncio-api-transaction` for more info.
+
+        Example:
+
+        .. code-block:: python
+
+            async for tx in con.retry():
+                with tx:
+                    value = tx.query_one("SELECT Counter.value")
+                    tx.execute(
+                        "UPDATE Counter SET { value := <int64>$value",
+                        value=value,
+                    )
+
+        Note executing queries on ``tx`` object rather than the original
+        connection.
+
+    .. py:method:: try_transaction()
+
+        Create an individual :py:class:`AsyncIOTransaction` object.
+
+        Transaction may fail so it's often better to use :py:meth:`retry`
+        instead. This method is targeted for users who want to implement \
+        more specialized transaction retry loop.
+
+        Example:
+
+        .. code-block:: python
+
+            with con.try_transaction() as tx:
+                value = tx.query_one("SELECT Counter.value")
+                tx.execute(
+                    "UPDATE Counter SET { value := <int64>$value",
+                    value=value,
+                )
+
+        Note executing queries on ``tx`` object rather than the original
+        connection.
 
     .. py:method:: transaction(isolation=None, readonly=None, deferrable=None)
+        **Deprecated**. Use :py:meth:`retry` or :py:meth:`try_transaction`.
 
         Create a :py:class:`AsyncIOTransaction` object.
 
@@ -261,89 +304,6 @@ Connection
     .. py:method:: is_closed()
 
         Return ``True`` if the connection is closed.
-
-
-.. _edgedb-python-asyncio-api-transaction:
-
-Transactions
-============
-
-The most common way to use transactions is through a context manager statement:
-
-.. code-block:: python
-
-   async with connection.transaction():
-       await connection.execute("INSERT User { name := 'Don' }")
-
-It is possible to nest transactions (a nested transaction context will create
-a savepoint):
-
-.. code-block:: python
-
-   async with connection.transaction():
-       await connection.execute(
-           'CREATE TYPE User { CREATE PROPERTY name -> str }')
-
-       try:
-           # Create a savepoint:
-           async with connection.transaction():
-               await connection.execute(
-                   "INSERT User { name := 'Don' }")
-               # This nested savepoint will be
-               # automatically rolled back:
-               raise Exception
-       except:
-           # Ignore exception
-           pass
-
-       # Because the nested savepoint was rolled back, there
-       # will be nothing in `User`.
-       assert (await connection.query('SELECT User')) == []
-
-Alternatively, transactions can be used without an ``async with`` block:
-
-.. code-block:: python
-
-    tr = connection.transaction()
-    await tr.start()
-    try:
-        ...
-    except:
-        await tr.rollback()
-        raise
-    else:
-        await tr.commit()
-
-
-See also the
-:py:meth:`AsyncIOConnection.transaction()` function.
-
-
-.. py:class:: AsyncIOTransaction
-
-    Represents a transaction or savepoint block.
-
-    Transactions are created by calling the
-    :py:meth:`AsyncIOConnection.transaction()` method.
-
-
-    .. py:coroutinemethod:: start()
-
-        Enter the transaction or savepoint block.
-
-    .. py:coroutinemethod:: commit()
-
-        Exit the transaction or savepoint block and commit changes.
-
-    .. py:coroutinemethod:: rollback()
-
-        Exit the transaction or savepoint block and discard changes.
-
-    .. describe:: async with c:
-
-        start and commit/rollback the transaction or savepoint block
-        automatically when entering and exiting the code inside the
-        context manager block.
 
 
 .. _edgedb-python-asyncio-api-pool:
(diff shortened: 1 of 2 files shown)


docs/api/asyncio_con.rst
```diff
@@ -234,8 +234,51 @@ Connection
             If the results of *query* are desired, :py:meth:`query` or
             :py:meth:`query_one` should be used instead.
 
+    .. py:method:: retry()
+
+        Create a :py:class:`AsyncIORetry` object. This is a way to make
+        reliable transactions. See
+        :ref:`edgedb-python-asyncio-api-transaction` for more info.
```
1st1, line 241: I'd make this description a bit more high level:

```suggestion
        Execute a retryable transaction.
        
        This is the preferred method of running transactions as it
        can auto-retry transactions in case a retryable [LINK] error (like
        a network disconnect) happens.
        
        It is very important to always keep in mind that the code within
        the ``retry()`` block might be executed more than one time.
        
        A ``retry()`` call returns an instance of
        :py:class:`AsyncIORetry`. 
        
        See :ref:`edgedb-python-asyncio-api-transaction` for more info.
```

We need to document which transactions are retryable in the main EdgeDB docs and link them from here.

I'd move the `See :ref:.. for more info` to the very bottom of this section, below the example.

docs/api/asyncio_con.rst
```diff
@@ -234,8 +234,51 @@ Connection
...
+        Note executing queries on ``tx`` object rather than the original
+        connection.
+
+    .. py:method:: try_transaction()
+
+        Create an individual :py:class:`AsyncIOTransaction` object.
+
+        Transaction may fail so it's often better to use :py:meth:`retry`
+        instead. This method is targeted for users who want to implement \
+        more specialized transaction retry loop.
```
1st1, line 264:
```suggestion

        Execute a non-retryable transaction.
        
        Contrary to ``retry()``, ``try_transaction()`` will not attempt
        to re-run the nested code block in case a retryable error happens.
        
        This is a low-level API and it is advised to use the ``retry()`` 
        method instead.

        A call to ``try_transaction()`` returns 
        :py:class:`AsyncIOTransaction`.
```

docs/api/asyncio_con.rst
1st1: cc @elprans 

Elvis take a look at how I phrased it. Any suggestions?

docs/api/asyncio_con.rst
```diff
@@ -387,15 +347,33 @@ Connection Pools
 
     :return: An instance of :py:class:`AsyncIOPool`.
 
-    Can be used either with an ``async with`` block:
+    Queries on the connection pool can be executed directly, in this
+    case every invocation some connection is picked for the duration of
+    the query:
```
1st1, line 352:
```suggestion
    The connection pool has high-level APIs to access Connection[link]
    APIs directly, without manually acquiring and releasing connections
    from the pool:
    
    * Pool.retry()
    * Pool.query()
    * Pool.query_one()
    * Pool.query_json()
    * Pool.query_one_json()
```

docs/api/asyncio_con.rst
```diff
@@ -387,15 +347,33 @@ Connection Pools
...
+    Transactions can be executed as well:
+
+    .. code-block:: python
+
+        async with edgedb.create_async_pool(user='edgedb') as pool:
+            async for tx in pool.retry():
+                async with tx:
+                    await tx.query('SELECT {1, 2, 3}')
+
+    If you need to lock on specific connection, use ``async with`` block:
```
1st1, line 368:
```suggestion
    To hold on a specific connection object, use the ``pool.acquire()`` API:
```

docs/api/asyncio_con.rst
```diff
@@ -550,3 +571,154 @@ Connection Pools
...
 
         Number of available connections in the pool.
+
+
+.. _edgedb-python-asyncio-api-transaction:
+
+Transactions
+============
+
+The most robust way to make transaction is using a retry loop:
```
1st1, line 581:
```suggestion
The most robust way to execute transactional code is to use 
the ``retry()`` API:
```

docs/api/asyncio_con.rst
```diff
@@ -550,3 +571,154 @@ Connection Pools
...
+
+.. code-block:: python
+
+    async for tx in pool.retry():
+        async with tx:
+            await tx.execute("INSERT User { name := 'Don' }")
+
+Note executing queries on ``tx`` object rather than original
+connection or connection pool.
```
1st1, line 590:
```suggestion
Note that we execute queries on the ``tx`` object in the above
example, rather than on the original connection pool ``pool`` 
object.
```

docs/api/asyncio_con.rst
```diff
@@ -550,3 +571,154 @@ Connection Pools
...
+
+This implies that:
+
+1. Transaction is executed atomically
+2. If transaction is failed for any of the number of transient errors (i.e.
+   network failure or concurrent update error) transaction is retried
+3. If any other exception occurs, transaction is rolled back and not
+   retried.
```
1st1, line 598:
```suggestion
The ``retry()`` API guarantees that:

1. Transactions are executed atomically;
2. If a transaction is failed for any of the number of transient errors (i.e.
   a network failure or a concurrent update error), the transaction would be retried;
3. If any other, non-retryable exception occurs, the transaction is rolled back,
   and the exception is propagated, immediately aborting the ``retry()`` block.
```

docs/api/asyncio_con.rst
```diff
@@ -550,3 +571,154 @@ Connection Pools
...
+
+One non-obvious implication of the (3) is that whole block is retried
```
1st1, line 600:
```suggestion
A non-obvious implication of retrying transactions is that the entire 
nested code block can be re-run, including any non-querying
Python code.
```

docs/api/asyncio_con.rst
```diff
@@ -550,3 +571,154 @@ Connection Pools
...
+                    UPDATE User FILTER .login = <str>$login
+                    SET { email_info := <json>$data}
+                ''',
+                login=login,
+                data=data,
+            )
+
+Here execution of the http request be retried too. The core of the issue
+is that whenever transaction is interrupted user might have changed the
+email, so we have to redo all the work done.
```
1st1, line 625:
```suggestion
In the above example, the execution of the HTTP request would be
retried too.
```

I don't understand the "The core of the issue..." sentence. How can the user change the email while the code is iterating with `async for` here?

docs/api/asyncio_con.rst
```diff
@@ -550,3 +571,154 @@ Connection Pools
...
+
+Generally it's recommended not to do potentially slow or long running
+external tasks within the transaction and split it into multiple
+ones.
```
1st1, line 629:
```suggestion
Generally it's recommended to not execute any long running
code within the transaction unless absolutely necessary.

Transactions allocate expensive server resources and having
too many concurrently running long-running transactions will
negatively impact the performance of the DB server.
```

"split it into multiple ones." -- this is not always possible. But I think we can just leave that sentence out now.

docs/api/asyncio_con.rst
```diff
@@ -550,3 +571,154 @@ Connection Pools
...
+* :py:meth:`AsyncIOConnection.retry()`
+* :py:meth:`AsyncIOConnection.try_transaction()`
+
+
+.. py:class:: AsyncIOTransaction
+
+    Represents a transaction or savepoint block.
+
+    Transactions are created by calling the
+    :py:meth:`AsyncIOConnection.transaction()` method.
```
1st1, line 644:
```suggestion
    Instances of this type are created by calling the
    :py:meth:`AsyncIOConnection.try_transaction()` method.
```

docs/api/asyncio_con.rst
```diff
@@ -550,3 +571,154 @@ Connection Pools
...
+* RFC1004_
+* :py:meth:`AsyncIOPool.retry()`
+* :py:meth:`AsyncIOPool.try_transaction()`
+* :py:meth:`AsyncIOConnection.retry()`
+* :py:meth:`AsyncIOConnection.try_transaction()`
+
+
+.. py:class:: AsyncIOTransaction
+
+    Represents a transaction or savepoint block.
```
1st1, line 641:
```suggestion
    Represents a transaction or a savepoint block.
```

docs/api/asyncio_con.rst
```diff
@@ -550,3 +571,154 @@ Connection Pools
...
+
+    Transactions are created by calling the
+    :py:meth:`AsyncIOConnection.transaction()` method.
+
+
+    .. py:coroutinemethod:: start()
+
+        Enter the transaction or savepoint block.
```
1st1, line 649:
```suggestion
        Start a transaction or create a savepoint.
```

1st1 (commented):
I've added some suggestions and comments.

docs/api/asyncio_con.rst
elprans: I prefer "retriable" instead of "retryable", because we say "transaction is retried", although the Internet seems to be split between the spellings.

As for the description, here's my take:

```suggestion
        Open a retriable transaction loop.
       
        This is the preferred method of initiating and running a database
        transaction in a robust fashion.  The `retry()` transaction loop will
        attempt to re-execute the transaction lop body if a transient error
        occurs, such as a network error or a transaction serialization error.
       
        Returns an instance of :py:class:`AsyncIORetry`.
       
        See :ref:`edgedb-python-asyncio-api-transaction` for more details.
```

docs/api/asyncio_con.rst
elprans: I think it's important to still use the word loop, so "the ``retry()`` loop API" would probably be a more informative description.  It is impossible to use "retry()" without a loop, isn't it?

docs/api/asyncio_con.rst
1st1: https://english.stackexchange.com/a/305274

elprans: That isn't an accepted answer, and I don't agree that "retriable" is necessarily a form of "triable", because, again, it's "retried", not "retryed"

1st1: Well, the "accepted" one isn't always the right one, since it's not accepted by some "expert", but rather by the one who asked the question in the first place :)  The accepted one doesn't speculate on which spelling is more correct either, so I posted a link to the comment that does.

docs/api/asyncio_con.rst
tailhook: Yes. Added "loop"

docs/api/asyncio_con.rst
tailhook: > I don't understand the "The core of the issue..." sentence. How can the user change the email while the code is iterating with async for here?

In parallel/concurrent request for example (or a background task or whatever).

I think we should mention somehow that we retry the whole transaction not because we're lazy, but because this is how transactions must be handled. Maybe we should pick a better example, where concurrent things are more expected to happen?

docs/api/asyncio_con.rst
1st1: @tailhook Let's merge Elvis' version, replacing retriable with retryable (as our native speakers suggested us to)

docs/api/asyncio_con.rst
```diff
@@ -387,15 +352,39 @@ Connection Pools
 
     :return: An instance of :py:class:`AsyncIOPool`.
 
-    Can be used either with an ``async with`` block:
+    The connection pool has high-level APIs to access Connection[link]
+    APIs directly, without manually acquiring and releasing connections
+    from the pool:
+
+    * Pool.retry()
```
1st1, line 359: Can we hyperlink these?

docs/api/asyncio_con.rst
```diff
@@ -539,6 +528,49 @@ Connection Pools
...
+            async for tx in pool.retry():
+                with tx:
+                    value = tx.query_one("SELECT Counter.value")
+                    tx.execute(
+                        "UPDATE Counter SET { value := <int64>$value",
+                        value=value,
+                    )
+
+        Note executing queries on ``tx`` object rather than the original
+        pool.
```
1st1, line 549:
```suggestion
        Note that we are executing queries on the ``tx`` object rather
        than on the original pool itself.
```

docs/api/asyncio_con.rst
```diff
@@ -539,6 +528,49 @@ Connection Pools
...
+
+            with pool.try_transaction() as tx:
+                value = tx.query_one("SELECT Counter.value")
+                tx.execute(
+                    "UPDATE Counter SET { value := <int64>$value",
+                    value=value,
+                )
+
+        Note executing queries on ``tx`` object rather than the original
+        pool.
```
1st1, line 572: ditto

docs/api/asyncio_con.rst
```diff
@@ -550,3 +582,160 @@ Connection Pools
...
+1. Transactions are executed atomically;
+2. If a transaction is failed for any of the number of transient errors (i.e.
+   a network failure or a concurrent update error), the transaction would be retried;
+3. If any other, non-retryable exception occurs, the transaction is rolled back,
+   and the exception is propagated, immediately aborting the ``retry()`` block.
+
+The key implication of retrying transactions is that the entire 
+nested code block can be re-run, including any non-querying
+Python code.
+including non-database statements, so for example:
```
1st1, line 616: dangling part of old sentence here

docs/api/asyncio_con.rst
1st1: Maybe rephrase to:

"In the above example, the execution of the HTTP request would be
retried too to guarantee that the DB state is synchronized with
the remote service state."

?

tailhook: Well, I this case remote service looks like readonly, so nothing synchronized. But generally I wanted to explain why we are retrying the whole transaction block rather than individual queries. Here it's a good idea to cache the value retrieved from server and only check the cache on the second retry, so actual request isn't something that has to be done.

docs/api/asyncio_con.rst
tailhook: Done

docs/api/asyncio_con.rst
tailhook: Fixed

docs/api/asyncio_con.rst
tailhook: Updated the first sentence and edited the text a bit: https://github.com/edgedb/edgedb-python/pull/150/commits/fc7dc2aa3e610a1f0d2fa1bf9ba88e4a4ce870b4

tailhook:
Any update on this? I think all the feedback was addressed except https://github.com/edgedb/edgedb-python/pull/150#pullrequestreview-577774084 which is not too bad I think and can be improved later.

docs/api/asyncio_con.rst
```diff
@@ -234,116 +234,88 @@ Connection
...
+            with con.try_transaction() as tx:
+                value = tx.query_one("SELECT Counter.value")
+                tx.execute(
+                    "UPDATE Counter SET { value := <int64>$value",
+                    value=value,
+                )
 
-.. py:class:: AsyncIOTransaction
+        Note executing queries on ``tx`` object rather than the original
+        connection.
```
1st1, line 290:
```suggestion
        Note that we are executing queries on the ``tx`` object,
        rather than on the original connection `con`.
```

docs/api/asyncio_con.rst
```diff
@@ -234,116 +234,88 @@ Connection
...
 
-    .. py:coroutinemethod:: start()
+        :param readonly:
+            Specifies whether or not this transaction is read-only.  If not
+            specified, the server-side default is used.
 
-        Enter the transaction or savepoint block.
+        :param deferrable:
+            Specifies whether or not this transaction is deferrable.  If not
+            specified, the server-side default is used.
```
1st1, line 308: I'd also link our transaction docs here that explain what readonly or deferrable do.

docs/api/asyncio_con.rst
```diff
@@ -387,15 +359,39 @@ Connection Pools
...
-    Can be used either with an ``async with`` block:
+    The connection pool has high-level APIs to access Connection[link]
+    APIs directly, without manually acquiring and releasing connections
+    from the pool:
+
+    * :py:meth:`AsyncIOPool.retry()`
+    * :py:meth:`AsyncIOPool.query()`
+    * :py:meth:`AsyncIOPool.query_one()`
+    * :py:meth:`AsyncIOPool.query_json()`
+    * :py:meth:`AsyncIOPool.query_one_json()`
```
1st1, line 370: Also `execute()`

docs/api/asyncio_con.rst
```diff
@@ -387,15 +359,39 @@ Connection Pools
...
+    Transactions can be executed as well:
+
+    .. code-block:: python
+
+        async with edgedb.create_async_pool(user='edgedb') as pool:
+            async for tx in pool.retry():
+                async with tx:
+                    await tx.query('SELECT {1, 2, 3}')
+
+    To hold on a specific connection object, use the ``pool.acquire()`` API:
```
1st1, line 386:
```suggestion
    To hold on to a specific connection object, use the ``pool.acquire()`` API:
```

docs/api/asyncio_con.rst
```diff
@@ -539,6 +535,51 @@ Connection Pools
         See :py:meth:`AsyncIOConnection.execute()
         <edgedb.AsyncIOConnection.execute>` for details.
 
+    .. py:method:: retry()
+
+        Create a :py:class:`AsyncIORetry` object. This is a way to make
```
1st1, line 540: Can we please copy the text from https://github.com/edgedb/edgedb-python/pull/150/files#diff-833af12d3f57fe855ba1925cc6b4409cef2094fb7aac456676721f0560e1e435R237?

docs/api/asyncio_con.rst
```diff
@@ -539,6 +535,51 @@ Connection Pools
...
+                        "UPDATE Counter SET { value := <int64>$value",
+                        value=value,
+                    )
+
+        Note that we are executing queries on the ``tx`` object rather
+        than on the original pool itself.
+
+    .. py:method:: try_transaction()
+
+        Create a :py:class:`AsyncIOTransaction` object.
```
1st1, line 561: Ditto, `Connection.try_transaction` has slightly better wording now.

docs/api/asyncio_con.rst
```diff
@@ -550,3 +591,161 @@ Connection Pools
...
         Number of available connections in the pool.
+
+
+.. _edgedb-python-asyncio-api-transaction:
+
+Transactions
+============
+
+The most robust way to execute transactional code is to use
+the ``retry()`` loop API:
```
1st1, line 602:
```suggestion
The most robust way to execute transactional code is to use
the ``retry()`` API:
```

1st1: Actually, never mind, let's keep "loop API". I think it better communicates to the reader that hey, this weird looking construct is actually the api.

docs/usage.rst
```diff
@@ -208,5 +176,53 @@ Below is an example of a connection pool usage:
...
+transactions.
+
 See :ref:`edgedb-python-asyncio-api-pool` API documentation for
 more information.
+
+
+Transactions
+------------
+
+The most robust way to create a transaction is ``retry`` method:
```
1st1, line 200: Link to the other, fuller explanation?

docs/api/asyncio_con.rst
```diff
@@ -550,3 +591,161 @@ Connection Pools
...
+
+        Exit the transaction or savepoint block and commit changes.
+
+    .. py:coroutinemethod:: rollback()
+
+        Exit the transaction or savepoint block and discard changes.
+
+    .. describe:: async with c:
+
+        start and commit/rollback the transaction or savepoint block
```
1st1, line 689:
```suggestion
        Start and commit/rollback the transaction or savepoint block
```

1st1 (approved):
I've left a few nit comments. Other than that it's good to go. Great job on this one!

docs/api/asyncio_con.rst
tailhook: This is deprecated part. Let's add this to the `with_transaction_options`.

PR #157, opened 2021-02-10
Fix `add_log_listener`

tailhook:
This is also a prerequisite for edgedb/edgedb#2190

### DIFF
--- a/tests/test_async_fetch.py
+++ b/tests/test_async_fetch.py
@@ -853,3 +853,24 @@ class TestAsyncFetch(tb.AsyncQueryTestCase):
             self.assertIsNot(conn_before, conn_after, "Reconnect expected")
         finally:
             await con.aclose()
+
+    async def test_async_log_message(self):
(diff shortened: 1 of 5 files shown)


tests/test_async_fetch.py
```diff
@@ -853,3 +853,24 @@ async def test_async_cancel_01(self):
...
+        msgs = []
+
+        def on_log(con, msg):
+            msgs.append(msg)
+
+        self.con.add_log_listener(on_log)
+        try:
+            await self.con.query(
+                'configure system set __internal_restart := true;')
+            await asyncio.sleep(0.01)  # allow the loop to call the callback
```
1st1, line 867: This might be a problem in CI -- sleeps like this are always unreliable. I'd better issue another query (no matter what) just to have another guaranteed roundtrip to the server during which the notice would be surely delivered.

1st1 (approved):

PR #159, opened 2021-02-11
Disable "transaction" capability when using `execute()`

tailhook:
Currently `test_async_wait_cancel_01` fails. Looks like this is a bug of edgedb.

### DIFF
--- a/edgedb/enums.py
+++ b/edgedb/enums.py
@@ -1,0 +1,14 @@
+import enum
+
+
+class Capability(enum.IntFlag):
+
+    MODIFICATIONS     = 1 << 0    # noqa
+    SESSION_CONFIG    = 1 << 1    # noqa
+    TRANSACTION       = 1 << 2    # noqa
+    DDL               = 1 << 3    # noqa
+    PERSISTENT_CONFIG = 1 << 4    # noqa
+
+
+Capability.ALL = 0xFFFF_FFFF_FFFF_FFFF
+Capability.EXECUTE = Capability.ALL & ~Capability.TRANSACTION
--- a/edgedb/_cluster.py
+++ b/edgedb/_cluster.py
@@ -196,7 +196,7 @@ class Cluster:
         conn_args['host'] = str(self._runstate_dir)
         conn_args['admin'] = True
         # startup of the database may take > 30 sec on github-actions
-        conn_args['wait_until_available'] = 120
+        conn_args['wait_until_available'] = 240
         conn = self.connect(**conn_args)
 
         try:
--- a/edgedb/asyncio_con.py
+++ b/edgedb/asyncio_con.py
@@ -31,6 +31,7 @@ from . import base_con
 from . import compat
 from . import con_utils
 from . import errors
+from . import enums
 from . import retry as _retry
 from . import transaction as _transaction
 from . import legacy_transaction
@@ -68,10 +69,12 @@ def _extract_errno(s):
 
 class _AsyncIOConnectionImpl:
 
-    def __init__(self):
+    def __init__(self, codecs_registry, query_cache):
         self._addr = None
         self._transport = None
         self._protocol = None
+        self._codecs_registry = codecs_registry
+        self._query_cache = query_cache
 
     def is_closed(self):
         transport = self._transport
@@ -158,8 +161,8 @@ class _AsyncIOConnectionImpl:
         self._protocol = pr
         self._addr = addr
 
-    async def execute(self, query):
-        await self._protocol.simple_query(query)
+    async def privileged_execute(self, query):
+        await self._protocol.simple_query(query, enums.Capability.ALL)
 
     def close(self):
         if self._protocol:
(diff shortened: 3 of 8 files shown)


edgedb/_taskgroup.py
```diff
@@ -210,6 +210,7 @@ def _abort(self):
 
         for t in self._tasks:
             if not t.done():
+                print("ABORT CANCEL")
```
1st1, line 213: debug

edgedb/_taskgroup.py
```diff
@@ -263,6 +264,7 @@ def _on_task_done(self, task):
             #        await something_else     # this line has to be called
             #                                 # after TaskGroup is finished.
             self._parent_cancel_requested = True
+            print("PARENT CANCEL")
```
1st1, line 267: ditto

edgedb/enums.py
```diff
@@ -0,0 +1,14 @@
...
+class Capability(enum.IntFlag):
+
+    MODIFICATIONS     = 1 << 0    # noqa
+    SESSION_CONFIG    = 1 << 1    # noqa
+    TRANSACTION       = 1 << 2    # noqa
+    DDL               = 1 << 3    # noqa
+    PERSISTENT_CONFIG = 1 << 4    # noqa
+
+
+Capability.ALL = 0xFFFFFFFFFFFFFFFF
```
1st1, line 13:
```suggestion
Capability.ALL = 0xFFFF_FFFF_FFFF_FFFF
```

1st1 (approved):
lgtm

tailhook:
Tests require #139

1st1:
Rebased onto the latest master. Let's see if the CI is happy now.

1st1:
Hm, the CI keeps failing on tests not able to connect to the cluster. @tailhook can you take a look?

tailhook:
So, it's unrelated to this fix (docs PR fails too: https://github.com/edgedb/edgedb-python/pull/156/checks?check_run_id=1927147357)

Also much larger timeouts looks like helpful here. But given variability of the time this tests runs (2m ... 20m) I think it's some bug in edgedb, with regard to connection handling :(

Alternatively it may be possible that this is race condition between pool tests and other tests (where pool tests suck up too much resources for doing 100 connection), but I don't think this is likely the reason.

1st1:
> Also much larger timeouts looks like helpful here. But given variability of the time this tests runs (2m ... 20m) ...

Why wasn't this an issue before?

> ... I think it's some bug in edgedb, with regard to connection handling :(

I really doubt that.

tailhook:
Looks like even edgedb-js has the [same issue](https://github.com/edgedb/edgedb-js/runs/1921801989) so most likely it's something with github actions. So should we merge this with increased timeouts?

tailhook:
> Looks like even edgedb-js has the [same issue](https://github.com/edgedb/edgedb-js/runs/1921801989) so most likely it's something with github actions. So should we merge this with increased timeouts?

Well, on the other hand, all of them depend on performance of edgedb, so it's still possible that edgedb has some recent issue.

1st1:
> Well, on the other hand, all of them depend on performance of edgedb, so it's still possible that edgedb has some recent issue.

The cost of establishing a new connection is pretty much zero now -- there are simply no awaits there (except message parsing which wasn't touched for a while). So I just don't see how we could have a regression there :/

elprans:
This isn't about new connections at all, it's server startup that gets bogged down massively.  We also use plain `unittest` here, which basically tries to start and init (or reuse) a cluster in `setUpClass` and there may be some wonkiness there.  We could extract `edb test` into a standalone package and use that instead for better reliability.

1st1:
> This isn't about new connections at all, it's server startup that gets bogged down massively. We also use plain unittest here, which basically tries to start and init (or reuse) a cluster in setUpClass and there may be some wonkiness there. We could extract edb test into a standalone package and use that instead for better reliability.

Hmm...

https://github.com/edgedb/edgedb-python/blob/7b2764881c1ec581810c0a9e5ef61182507bbd11/edgedb/_testbase.py#L234-L237

Looks like `_start_cluster` doesn't block until the cluster has started and can be connected to (unlike in edgedb/edgedb). We should fix that.

1st1:
@elprans 

> Looks like even edgedb-js has the same issue so most likely it's something with github actions. So should we merge this with increased timeouts?

This is pretty weird then -- the JS CI waits until the server is fully available and is accepting connections before starting to test it. Why wasn't 5 seconds timeout enough here?

elprans:
It's hard to say without seeing the server output.

tailhook:
The problem with blaming startup is this:
https://github.com/edgedb/edgedb-python/runs/1927543605

After I increased `wait_until_available` 2x to 4 min, this is what started to pop up. Note: last failure is not `wait_until_available` timeout it's a connect timeout. So after that startup thing was already able to connect, connection pool is still slow (at least > 10 sec for individual connection, and I think python tests aren't parallelized, right?)

elprans:
> The problem with blaming startup is this:

OK, there the problem _is_ the connection cost, but then it's a7.  Once b1 is out, things will be fast both for `stable` and for `nightly`.

1st1:
Alright, I'm gonna take a bit closer look at this now.

1st1 (approved):

PR #161, opened 2021-02-12
Reconnect on more conditions

tailhook:
Previously we requires OSError with errno as a condition for reconnect.
But this commit allows toreconnect on manually thrown errors such as
`ConnectionAbortedError`, and also ones thrown protocol `connect()`
method (not just on socket connect method).


tailhook:
The test that fails is unrelated to this PR.

elprans (approved):

Issue #160, opened 2021-02-11
`_advisory_lock` tests fails


tailhook:
EdgeDB: 1.0a8.dev5422+g7d7d371f7

Log: https://github.com/edgedb/edgedb-python/pull/139/checks?check_run_id=1881955387#step:7:328

PR #162, opened 2021-02-12
Fixes for tests

elprans:
There are a number of issues:

- `asyncio.wait_for()` is very buggy in Python 3.6, so backport
  fixes from newer Pythons as a wrapper and use it everywhere;
- There was a single `sys::sleep` user which wasn't adjusted
  to use `sys::_sleep` (and also lacked a skip guard);
- The `test_async_wait_cancel_01` test had a bug in it: it used the
  wrong connection instance.

Fixes: #160.

### DIFF
--- a/edgedb/_cluster.py
+++ b/edgedb/_cluster.py
@@ -94,7 +94,7 @@ class Cluster:
 
     @staticmethod
     def get_edgedb_server():
-        return 'edgedb-server'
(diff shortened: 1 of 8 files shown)


tailhook (approved):
Generally looks good. Just few questions.

Also do tests crash because of uvloop 0.15?

PR #163, opened 2021-02-17
No `uvloop` for tests

tailhook:
Looks like previously was unused

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ from setuptools.command import build_py as setuptools_build_py
 from setuptools.command import sdist as setuptools_sdist
 
 
-CYTHON_DEPENDENCY = 'Cython==0.29.21'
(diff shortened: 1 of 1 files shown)


elprans (approved):

setup.py
```diff
@@ -49,7 +49,6 @@
     'pycodestyle~=2.6.0',
     'pyflakes~=2.2.0',
     'flake8~=3.8.1',
-    'uvloop>=0.15.1;platform_system!="Windows"',
```
1st1, line 52: Can we actually keep it and install it on Python > 3.6? Can we add one more condition here? Alternatively we can make the GH CI Action to install it and then run tests with it.

1st1:
Superseded by https://github.com/edgedb/edgedb-python/pull/164

PR #164, opened 2021-02-18
Test with uvloop on Python >= 3.7

1st1:


.github/workflows/tests.yml
```diff
@@ -61,6 +61,11 @@ jobs:
       run: |
         python setup.py test
 
+    - name: Test under uvloop
```
elprans, line 64: I'd parallelize this into a job via a `loop: [asyncio, uvloop]` matrix

1st1: done

elprans (approved):

PR #165, opened 2021-02-19
Rely on edgedb itself to find a suitable test port to run on

1st1:


edgedb/_cluster.py
```diff
@@ -136,29 +109,54 @@ def init(self):
...
-            env=env, cwd=str(self._data_dir))
+            stdout=stdout,
+            stderr=subprocess.DEVNULL,
+            env=env,
+            universal_newlines=True,
+            cwd=str(self._data_dir)
+        )
+
+        if port == 0:
+            while 1:
```
tailhook, line 148: I would prefer there be two things:
1. Timeout
2. Spit edgedb output when timeout happens or when `json.loads` fails (including stderr). This was already a problem in JS (https://github.com/edgedb/edgedb-js/commit/3bdb7482bb4a9613c7f19e7fa3a162561d84b311)

tailhook (changes requested):

PR #166, opened 2021-02-20
Make sure aclose() waits for the server to close the connection

elprans:
The point of the `aclose()` and the reason why it's a coroutine is to wait
for the server to close the connection.  This is known as graceful
disconnect, as opposed to a fast disconnect via `terminate()`.

Graceful disconnects are especially useful for cases when an agreement
on the state of the server after the disconnect is important, such as
when the script runs `DROP DATABASE` immediately after disconnecting.
If the disconnect is not negotiated properly, there is a possibility
that the server still thinks the connection is open and so a `DROP
DATABASE` operation would fail complaining that the database is still
being accessed.

### DIFF
--- a/edgedb/asyncio_con.py
+++ b/edgedb/asyncio_con.py
@@ -51,12 +51,8 @@ class _AsyncIOConnectionImpl:
         self._query_cache = query_cache
 
     def is_closed(self):
-        transport = self._transport
         protocol = self._protocol
-        return not (
-            transport and not transport.is_closing() and
-            protocol and protocol.connected
-        )
+        return protocol is None or not protocol.connected
 
     async def connect(self, loop, addrs, config, params, *,
                       single_attempt=False, connection):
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -29,7 +29,6 @@ _touch:
 	rm -fr $(ROOT)/edgedb/protocol/protocol.c
 	find $(ROOT)/edgedb/protocol -name '*.pyx' | xargs touch
 	find $(ROOT)/edgedb/datatypes -name '*.pyx' | xargs touch
-	find $(ROOT)/edgedb/datatypes -name '*.c' | xargs touch
```
1st1, line 32: Is this supposed to be part of this commit?

1st1 (approved):

Makefile
elprans: Not really, but that line is questionable as it cancels the effect of the preceding .pyx touch.

PR #167, opened 2021-02-20
Simplify test workflow

elprans:
Use one step to test with a particular loop implementation and use
`exclude` to black out unsupported matrix variants.


1st1 (approved):

1st1:
@elprans can you rebase?

PR #174, opened 2021-03-12
Implement `with_transaction_options` and `with_retry_options` (RFC1004)

tailhook:
This omits docs, and only includes docstrings for few most important things (mostly so reviewer could figure out semantics more easily). I'll file another PR for documentation and docstrings of more things.

### DIFF
--- a/edgedb/options.py
+++ b/edgedb/options.py
@@ -15,7 +15,7 @@ def default_backoff(attempt):
 
 class RetryCondition:
     """Specific condition to retry on for fine-grained control"""
-    Conflict = enum.auto()
+    TransactionConflict = enum.auto()
     NetworkError = enum.auto()
 
 
@@ -62,9 +62,9 @@ class RetryOptions:
         res = default
         if overrides:
             if isinstance(exception, errors.TransactionConflictError):
-                res = overrides.get(RetryCondition.Conflict, default)
+                res = overrides.get(RetryCondition.TransactionConflict, res)
             elif isinstance(exception, errors.ClientError):
-                res = overrides.get(RetryCondition.NetworkError, default)
+                res = overrides.get(RetryCondition.NetworkError, res)
         return res
 
 
--- a/edgedb/options.py
+++ b/edgedb/options.py
@@ -72,7 +72,7 @@ class RetryOptions:
 
 
 class TransactionOptions:
-    """Options for `raw_transaction()` an `retrying_transaction()`"""
+    """Options for `raw_transaction()` and `retrying_transaction()`"""
     __slots__ = ['_isolation', '_readonly', '_deferrable']
 
     def __init__(
--- a/edgedb/options.py
+++ b/edgedb/options.py
@@ -22,8 +22,8 @@ class RetryCondition:
 
 class IsolationLevel:
     """Isolation level for transaction"""
-    Serializable = enum.auto()
-    RepeatableRead = enum.auto()
+    Serializable = "SERIALIZABLE"
+    RepeatableRead = "REPEATABLE READ"
 
 
 class RetryOptions:
@@ -89,6 +89,28 @@ class TransactionOptions:
     def defaults(cls):
         return cls()
 
+    def start_transaction_query(self):
+        isolation = str(self._isolation)
+        if self._readonly:
+            mode = 'READ ONLY'
+        else:
+            mode = 'READ WRITE'
+
+        if self._deferrable:
+            defer = 'DEFERRABLE'
+        else:
+            defer = 'NOT DEFERRABLE'
+
+        return f'START TRANSACTION ISOLATION {isolation}, {mode}, {defer};'
+
+    def __repr__(self):
+        return (
+            f'<{self.__class__.__name__} '
+            f'isolation:{self._isolation}, '
+            f'readonly:{self._readonly}, '
+            f'deferrable:{self._deferrable}>'
+        )
+
 
 class _OptionsMixin:
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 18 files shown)


edgedb/options.py
```diff
@@ -0,0 +1,291 @@
...
+                res = overrides.get(RetryCondition.SerializationError, default)
+            elif isinstance(exception, errors.TansactionDeadlockError):
+                res = overrides.get(RetryCondition.Deadlock, default)
+            elif isinstance(exception, errors.ClientError):
+                res = overrides.get(RetryCondition.NetworkError, default)
+        return res
+
+
+class TransactionOptions:
+    """Options for `raw_transaction()` an `retrying_transaction()`"""
```
1st1, line 75:
```suggestion
    """Options for `raw_transaction()` and `retrying_transaction()`"""
```

tests/test_sync_tx.py
```diff
@@ -62,8 +63,42 @@ def test_sync_transaction_regular_01(self):
...
+            for tx in con.retrying_transaction():
+                with tx:
+                    pass
+            if not any(v is None for v in (isolation, readonly, deferrable)):
+                options = edgedb.TransactionOptions(
+                    isolation=isolation,
+                    readonly=readonly,
+                    deferrable=deferrable,
+                )
+                con = self.con.with_transaction_options(options)
```
1st1, line 93: I think this API is very verbose. IMO

```python
self.con.with_transaction_options(
   isolation=....
)
```

would be way more convenient. The need to create `edgedb.TransactionOptions` object manually reminds me of early Java APIs a lot.

edgedb/options.py
```diff
@@ -0,0 +1,291 @@
...
+
+    @property
+    def retry_options(self):
+        return self._retry_options
+
+    @property
+    def transaction_options(self):
+        return self._transaction_options
+
+    def with_retry_options(self, options: RetryOptions):
```
1st1, line 274: I don't see tests for this method. But as I mentioned in another comment, the need to create `RetryOptions` object separately makes the API extremely verbose and make it feel low level.

1st1:
@elprans see my comments for this PR, I'd appreciate you taking a look when you have a moment.

tests/test_sync_tx.py
1st1: FWIW https://github.com/edgedb/rfcs/blob/master/text/1004-transactions-api.rst defines the method as `async def with_transaction_options(self, isolation: ...)` -- i.e. it receives keyword arguments and doesn't require a public `edgedb.TransactionOptions` type.  While the RFC hasn't yet been set in stone (we can tweak it while we're implementing it) I don't see why this departure from it was necessary.

edgedb/options.py
tailhook: This is not a user-visible method. So it get's normalized `RetryOptions` object.

tests/test_sync_tx.py
tailhook: You can either use:
```
with_transaction_options(
  isolation=, deferred=..)
```
Or having encapsulated object:
```
con.with_transaction_options(options)
```
Here we use latter because we reuse `options` object multiple times.

edgedb/options.py
```diff
@@ -0,0 +1,291 @@
...
+        )
+        result = RetryOptions.__new__(RetryOptions)
+        result._default = default
+        result._overrides = overrides
+        return result
+
+    @classmethod
+    def defaults(cls):
+        return cls(
+            attempts=3,
```
fmoor, line 56: Does attempts mean that 3 total attempts will be made or that 3 additional attempts will be made after an initial attempt?

tailhook: 3 in total. So one initial and two additional in case of error.

tests/test_sync_tx.py
1st1: > You can either use:

Yeah, I see that, but why? I'd prefer us to only have keyword-only args, using a separate "options" object here seems like API bloat.

edgedb/options.py
1st1: Well, let's move my comment to line 172 ;) The question still stands: `_OptionsMixin.with_retry_options` seems to lack tests, no?

tests/test_sync_tx.py
tailhook: So for `with_retry_options`. If we want to have some complex configuration, we don't have a way to store it efificently, we have to do something like that:
```python
def my_options1(obj):
   obj = obj.with_retry_options(attempts=3, backoff=backoff1)
   obj = obj.with_retry_options(RetryCondition.Deadlock, attempts=2, backoff=backoff2)
   return obj
```
And then use the function instead of `with_retry_options` in code. Alternatively you may keep list of tuples (and make sure that default option is the first) which isn't super convenient either (and requires another utility function anyway).

So instead of doing that you can have an initialized `RetryOptions` object and use regular `with_retry_options`.

Now, if you look at `with_transaction_options` you may think that it's simpler (a single dict) so you don't ever need `TransactionOptions`. But the issue is forward compatibility. Say if we add some options, like `transaction_timeout`.

Then using this example, you make sure that you replace all options, so `transaction_timeout` has default value:
```python
opt = TransactionOptions(isolated=i, deferred=d, readonly=r)
con1 = con.with_transaction_options(opt)
```
But with this code, you only override some options:
```python
con1 = con.with_transaction_options(isolated=i, deferred=d, readonly=r)
```

So in general keyword-arguments update only selected option(s) for convenience. And positional arguments make sure that configuration is exactly the same as you specified, regardless of future additions.

edgedb/options.py
tailhook: Yes, no tests options setters. I'll add when we decide that current approach is okay.

tests/test_sync_tx.py
elprans: > `   obj = obj.with_retry_options(RetryCondition.Deadlock, attempts=2, backoff=backoff2)`

This seems to me like too much configurability.  Is it really necessary to specify retry policy with so much granularity?

> So in general keyword-arguments update only selected option(s) for convenience. And positional arguments make sure that configuration is exactly the same as you specified, regardless of future additions.

I think this is confusing and unexpected, since the method changes its behavior in a major way depending on how you call it.  IMO, the *addition* and *complete replacement* of options should be two distinct operations.

It's also unclear to me if there is a real benefit of adding `RetryOptions` to the driver as opposed to leaving it to user code.  If your configuration is so complex, chances are you wouldn't mind rolling your own wrapper around the kwargs dict (or just use dicts in the first place).

Finally, the paragraph above makes me think that if we do decide to add a standalone `Options` object, it should be a mapping type, so instead of `with_transaction_options(opts)` you do `with_transaction_options(**opts)`, which partly addresses my point about argument confusion by removing an extra calling convention.

tailhook: I see this differently:
1. Granular configuration was defined in RFC (the actual definition is a little bit refined here, but this is not the core issue). Yes I really think that errors different by their nature (i.e. network error and deadlock) should often be treated differently.
2. I think that configuration here will be refined later (one example: one might want to adjust number of retries based on metrics, i.e. less retries if system is under huge load -- we add a callback to the number of retries)
3. I don't think it makes sense to have `with_replace_transaction_options` and `with_update_transaction_options`. The behavior of methods as are stated currently is very clear: there is no "partial" `TransactionOptions` type, so it's clear that all options are got from the object, conversely it's uncommon to have *required* keyword arguments, so selectively changing options using kwargs is quite common.
4. We already have precedent of kwargs becoming standalone builder object: connection arguments. So this idea is not local to just these two methods.
5. I always find it awkward when I have to invent my own types for storing some library's options. Even more: current types are immutable and have nice modification methods. Reinventing those is a substantial effort in many languages. Having them reused in multiple connections (even across the user code) is a nice bonus.

So if you want to simplify things a bit, I think we should go the other way, we may remove `condition` from `with_retry_options` and only allow fine-grained control using `RetryOptions.with_rule`. I don't see the value in dropping fine-grained control (unless our conditions are not well-defined enough).

Forward compatibility matters.

tailhook: Another examples of future extensions:
1. `TransactionOptions.with_replica_options(max_latency=10)`
2. `ReplyOptions.with_replica_policy(RetryOnDifferentReplica)`

These might be harder to fit to the dictionary nicely (e.g. `with_replica_options` probably sets `readonly` by default).

Also with more params the cost of unpacking the dict will grow.

edgedb/options.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+def default_backoff(attempt):
+    return (2 ** attempt) * 0.1 + random.randrange(100) * 0.001
+
+
+class RetryCondition:
+    """Specific condition to retry on for fine-grained control"""
+    SerializationError = enum.auto()
+    Deadlock = enum.auto()
```
elprans, line 19: Merge this into `TransactionConflictError` as discussed in edgedb/edgedb-go#123

elprans (approved):
As discussed, let's get rid of kwargs-based incremental update for `TransactionOptions` and `RetryOptions`.

Other than that, LGTM! Great work, @tailhook!

1st1:
Please ping me when the new stuff is in. I'll take another quick look.

edgedb/options.py
tailhook: Added tests for retry with different options.

tailhook:
@1st1 should be ready now.

edgedb/options.py
elprans: `Conflict` is a bit too generic.  Can we say `TransactionConflict`?

tailhook: Well, yes, since we have `UNLESS CONFLICT` one may imply that this is related.

1st1 (approved):

tailhook:
Waits for #177

PR #175, opened 2021-03-28
fix: use setuptools as distutils is deprecated

kumaraditya303:


setup.py
```diff
@@ -17,27 +17,17 @@
 #
 
 
-import sys
```
kumaraditya303, line 20: Not required as pip by default shows this error

setup.py
```diff
@@ -17,27 +17,17 @@
...
-    # edgedb.NamedTuple() relies on ordered **kwargs.
-    raise RuntimeError('edgedb requires Python 3.6 or greater')
-
 import os
-import os.path
 import pathlib
 import re
 import subprocess
-
-# We use vanilla build_ext, to avoid importing Cython via
```
kumaraditya303, line 32: `distutils` is deprecated and will be removed in next version https://www.python.org/dev/peps/pep-0632/

setup.py
```diff
@@ -214,31 +205,10 @@ def finalize_options(self):
                         need_cythonize = True
 
         if need_cythonize:
-            import pkg_resources
```
kumaraditya303, line 217: `setuptools` does this fine without this

setup.py
```diff
@@ -291,56 +250,60 @@ def finalize_options(self):
...
             extra_link_args=LDFLAGS,
-            include_dirs=INCLUDE_DIRS),
+            include_dirs=INCLUDE_DIRS,
+        ),
     ],
-    cmdclass={'build_ext': build_ext},
+    cmdclass={'sdist': sdist, 'build_py': build_py, 'build_ext': build_ext},
     test_suite='tests.suite',
     install_requires=[
-        'typing-extensions~=3.7.4',
```
kumaraditya303, line 342: Only install `typing-extensions` on python < 3.9 as the package only uses `TypedDict` and `Literal` which are present in python >3.9

setup.py
kumaraditya303:
```console
ERROR: Package 'edgedb' requires a different Python: 3.5.1 not in '>=3.7'
```

setup.py
```diff
@@ -75,29 +67,29 @@
...
 
 if SYSTEM == 'darwin':
     # Lots of warnings from the standard library on macOS 10.14
     CFLAGS.extend(['-Wno-nullability-completeness'])
 
-_ROOT = pathlib.Path(__file__).parent
+_ROOT = pathlib.Path(__file__).parent.resolve()
 
 
-with open(str(_ROOT / 'README.rst')) as f:
```
kumaraditya303, line 88: Conversion of `Path` to `str` is not required

setup.py
```diff
@@ -122,31 +115,28 @@ def git_commitish():
 
 
 class VersionMixin:
```
kumaraditya303, line 117: `setuptools_scm` can be used to create version from git

1st1:
@fantix can you take a look at this when you have a moment?

fantix:
Sure thing!

kumaraditya303:
@fantix Not sure why the tests are failing because of github token as I can see.

setup.py
```diff
@@ -263,23 +234,11 @@ def finalize_options(self):
...
-setup_requires = []
-
-if (not (_ROOT / 'edgedb' / 'protocol' / 'protocol.c').exists() or
-        '--cython-always' in sys.argv):
-    # No Cython output, require Cython to build.
-    setup_requires.append(CYTHON_DEPENDENCY)
-
-
-with open(str(_ROOT / 'README.rst')) as f:
-    readme = f.read()
```
fantix, line 275: This was a feature that when installing from sdist tarballs, Cython is NOT installed (temporarily and then removed) as we deliver the `.c` files too.

setup.py
```diff
@@ -291,56 +250,60 @@ def finalize_options(self):
...
+            'edgedb.protocol.blocking_proto',
+            ['edgedb/protocol/blocking_proto.pyx'],
             extra_compile_args=CFLAGS,
             extra_link_args=LDFLAGS,
-            include_dirs=INCLUDE_DIRS),
+            include_dirs=INCLUDE_DIRS,
+        ),
     ],
-    cmdclass={'build_ext': build_ext},
+    cmdclass={'sdist': sdist, 'build_py': build_py, 'build_ext': build_ext},
```
fantix, line 301: Oops, nice catch!

fantix (changes requested):
Thanks for the PR! Please separate the reformatting into a separate commit - that would really help with history tracing and further code review.

The failing test is I think a bug in the workflow - I'll try to fix that tmr.

setup.py
```diff
@@ -291,56 +250,60 @@ def finalize_options(self):
...
     test_suite='tests.suite',
     install_requires=[
-        'typing-extensions~=3.7.4',
+        'typing-extensions~=3.7.4;python_version<="3.8"',
     ],
+    python_requires='>=3.7',
```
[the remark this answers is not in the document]
kumaraditya303: Ok I will separate the formatting, my editor auto formatted with black the whole file 😅

setup.py
```diff
@@ -75,29 +67,29 @@
 SYSTEM = sys.platform
 
 if SYSTEM != 'win32':
-    CFLAGS.extend(['-std=gnu99', '-fsigned-char', '-Wall',
-                   '-Wsign-compare', '-Wconversion'])
+    CFLAGS.extend(
+        ['-std=gnu99', '-fsigned-char', '-Wall', '-Wsign-compare', '-Wconversion']
+    )
```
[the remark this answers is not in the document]
kumaraditya303: Black uses double quotes, is that okay ?

setup.py
kumaraditya303: Got it. It would be nice if this project used PEP 570 build system as would suggest @fantix

setup.py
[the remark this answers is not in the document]
fantix: Sorry for the late reply - that's not okay. New code could optionally follow the `black -S -l 79` format, but we don't want to do a complete reformatting yet.

PR #180, opened 2021-05-05
Implement reading project config

tailhook:
Details:
1. Uses current directory for project dir
2. Does not search of the ancestor directories (unlike CLI tool)
3. Now crashes if `edgedb.toml` not found and no parameters specified

Also includes implementation of `EDGEDB_INSTANCE` env var support

### DIFF
--- a/edgedb/con_utils.py
+++ b/edgedb/con_utils.py
@@ -25,6 +25,7 @@ import typing
 import urllib.parse
 import warnings
 import pathlib
+import hashlib
 
 from . import errors
 from . import credentials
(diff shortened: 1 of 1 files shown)


1st1 (approved):
Looks good. Please implement the suggested error wording from the similar PR in edgedb-js. And just like in edgedb-js, I wish we could have a test for this.

edgedb/con_utils.py
```diff
@@ -120,6 +121,14 @@ def _parse_hostlist(hostlist, port):
     return hosts, port
 
 
+def _stash_path(path):
+    path = os.path.realpath(path)
```
jaclarke, line 125:
```suggestion
    path = os.path.realpath(path)
    if _system == 'Windows' and not path.startswith('\\\\'):
        path = '\\\\?\\' + path
```

1st1:
@tailhook Do you have time to finalize this one? Would love to begin the release process soon

tailhook:
Merged, suggestion it seems to work. But looks like I can't make a test today. Stuck with JS test and now with merging deno.

PR #184, opened 2021-05-13
add windows to github CI

fmoor:
Run test on windows as well as linux and macos.


edgedb/_testbase.py
```diff
@@ -52,52 +52,58 @@ def flt(log_record):
...
+        raise _default_cluster
+
+    if _default_cluster:
+        return _default_cluster
+
+    args = [
+        os.environ.get('EDGEDB_SERVER_BINARY', 'edgedb-server'),
+        "--temp-dir",
+        "--testmode",
+        "--echo-runtime-info",
```
elprans, line 71: Use `--emit-server-status=` instead.  That way you'd be able to leave stdin/stderr undisturbed, which, in turn, will simplify debug.

edgedb/_testbase.py
```diff
@@ -52,54 +54,74 @@ def flt(log_record):
...
+            "--temp-dir",
+            "--testmode",
+            f"--emit-server-status={status_file}",
+            "--port=auto",
+            "--bootstrap-command=ALTER ROLE edgedb { SET password := 'test' }",
+        ]
+        cwd = '/tmp'
+
+        if sys.platform == 'win32':
+            args = ['wsl', 'sudo', '-u', 'edgedb'] + args
```
elprans, line 85: `wsl.exe` supports the `-u` flag directly, you don't need to do `sudo`, I think.

edgedb/_testbase.py
```diff
@@ -52,54 +54,74 @@ def flt(log_record):
...
+
+        env = os.environ.copy()
+        # Make sure the PYTHONPATH of _this_ process does
+        # not interfere with the server's.
+        env.pop('PYTHONPATH', None)
+
+        p = subprocess.Popen(args, env=env, cwd=cwd)
+
+        if sys.platform == 'win32':
+            status_file = pathlib.Path(r'C:\Temp\edgedb-server-status.txt')
```
elprans, line 96: `C:\Temp` might not be writable, use `tempfile.TemporaryDirectory` instead.

edgedb/_testbase.py
```diff
@@ -52,54 +54,74 @@ def flt(log_record):
...
+            print('iteration number:', i)
+            if status_file.exists():
+                break
+            time.sleep(1)
+        else:
+            raise RuntimeError('server status file not found')
 
-    if cleanup_atexit:
-        atexit.register(_shutdown_cluster, cluster, destroy=destroy)
+        with status_file.open('rb') as f:
```
elprans, line 106: I'd put the reading of the file inside the retry loop as well, because otherwise you might get a race between the file getting created and data being flushed into it.  Basically: 1) try opening the file; 2) iterate over lines and look for `READY=`; 3) if found, break.

edgedb/_testbase.py
```diff
@@ -52,54 +54,70 @@ def flt(log_record):
...
+            os.environ.get('EDGEDB_SERVER_BINARY', 'edgedb-server'),
+            "--temp-dir",
+            "--testmode",
+            f"--emit-server-status={status_file}",
+            "--port=auto",
+            "--bootstrap-command=ALTER ROLE edgedb { SET password := 'test' }",
+        ]
+
+        if sys.platform == 'win32':
+            args = ['wsl', '-u', 'nobody'] + args
```
elprans, line 85: Why not `edgedb`?

fmoor: changed

edgedb/_testbase.py
fmoor: changed

edgedb/_testbase.py
fmoor: changed

edgedb/_testbase.py
fmoor: changed

elprans:
@fmoor, please rebase.

.github/workflows/tests.yml
```diff
@@ -76,8 +76,8 @@ jobs:
     - name: Install Python Deps
       if: steps.release.outputs.version == 0
       run: |
-        pip install --upgrade setuptools pip wheel
-        pip install -e .[test]
+        pip install --user --upgrade setuptools pip wheel
```
elprans, line 79: This isn't necessary, as there is an implicit virtualenv.  If you're trying to fix a `EPERM` on Windows, then the correct fix is replacing `pip` with `python -m pip`

edgedb/_testbase.py
```diff
@@ -52,52 +54,76 @@ def flt(log_record):
...
+            .lower()
+        )
+
+        args = [
+            os.environ.get('EDGEDB_SERVER_BINARY', 'edgedb-server'),
+            "--temp-dir",
+            "--testmode",
+            f"--emit-server-status={status_file_unix}",
+            "--port=auto",
+            "--bootstrap-command=ALTER ROLE edgedb { SET password := 'test' }",
```
elprans, line 86: Is there a reason why there isn't `--auto-shutdown` here?

edgedb/_testbase.py
```diff
@@ -52,52 +54,76 @@ def flt(log_record):
...
+    if _default_cluster:
+        return _default_cluster
+
+    try:
+        tmpdir = tempfile.TemporaryDirectory()
+        status_file = os.path.join(tmpdir.name, 'server-status')
+
+        # if running on windows adjust the path for WSL
+        status_file_unix = (
+            status_file.replace('C:', '/mnt/c')
```
elprans, line 75: It could be other drive here, so I'd go for a regexp replacement.

edgedb/_testbase.py
fmoor: The test suite currently creates a connection per `TestCase` class. With `--auto-shutdown` the server exits after the first `TestCase` closes it's connection and the rest of the tests fail.

.github/workflows/tests.yml
fmoor: changed

edgedb/_testbase.py
elprans: Hmm. I don't see how the server is being shut down then.  I think we should modify `--auto-shutdown` to accept a timeout.

fmoor: I added a global connection and `atexit.register(con.close)` so that there is a persistent connection throughout the test run.

fmoor: and added `--auto-shutdown`

elprans: OK. Still, let's implement shutdown delay in the server.  I think something like 5 seconds by default should address the issue and we can make it configurable via a new `--auto-shutdown-after=Ns` option or something like that.

edgedb/_testbase.py
```diff
@@ -52,52 +55,80 @@ def flt(log_record):
...
+    if _default_cluster:
+        return _default_cluster
 
+    try:
+        tmpdir = tempfile.TemporaryDirectory()
+        status_file = os.path.join(tmpdir.name, 'server-status')
+
+        # if running on windows adjust the path for WSL
+        status_file_unix = (
+            re.sub(r'$[a-zA-Z]:', '/mnt/c', status_file)
```
elprans, line 76:
```suggestion
            re.sub(r'^([a-zA-Z]):', lambda m: f'{m.group(1).lower()}/mnt/', status_file)
```

fmoor: changed

elprans (approved):

edgedb/_testbase.py
fmoor: changed

edgedb/_testbase.py
fmoor: https://github.com/edgedb/edgedb/issues/2565

PR #189, opened 2021-06-04
Use HTTP Upgrade to switch to binary protocol

fantix:


setup.py
```diff
@@ -339,6 +339,7 @@ def finalize_options(self):
     cmdclass={'build_ext': build_ext},
     test_suite='tests.suite',
     install_requires=[
+        'httptools~=0.2',
```
1st1, line 342: Oh, I'm not sure I want to add a dependency. Perhaps we should revisit the HTTP upgrade idea -- keep it for those who want to implement it, but allow the current hack of just detecting the binary protocol signature.

nsidnev: Hi! Could you please share a link that will explain why this change in drivers and EdgeDB is necessary? I looked in `edgedb/rfcs` but couldn't find an explanation there, maybe I missed something there?

jaclarke: https://github.com/edgedb/edgedb/discussions/1910

1st1: @nsidnev a quick heads up: we'll be soon switching to always enabled TLS for the binary protocol. Which means that we'll be going away from HTTP1.1 upgrade, using TLS ALPN to upgrade the protocol instead. We'll have more details soon. But I wanted to mention this to you so that you don't implement the HTTP upgrade in your binding.

PR #193, opened 2021-06-08
Add flake8-bugbear to test requirements

elprans:
We use flake8-bugbear in the server codebase, so add this here too for
consistency.


1st1 (approved):
The commit message is wrong; pytest -> flake8

PR #195, opened 2021-06-18
support binary protocol version 0.11

fmoor:
Increases the number of bytes dedicated to the object descriptor element
flag field from 8 to 32 bytes.

server change https://github.com/edgedb/edgedb/pull/2603


edgedb/protocol/protocol.pyx
```diff
@@ -132,7 +132,7 @@ cdef class SansIOProtocol:
 
         self.server_settings = {}
         self.reset_status()
-        self.protocol_version = (PROTO_VER_MAJOR, 0)
+        self.protocol_version = (PROTO_VER_MAJOR, PROTO_VER_MINOR)
```
tailhook, line 135: I think this is bad to assume highest protocol version before negotiation is done. And I don't see any need to change it in this PR.

fmoor: changed

tests/test_datetime.py
```diff
@@ -62,8 +62,11 @@ async def test_duration_01(self):
         ''', durs)
         self.assertEqual(list(durs_from_db), durs)
 
-    @unittest.expectedFailure
     async def test_relative_duration_01(self):
+        try:
+            self.con.query("SELECT <cal::relative_duration>'1y'")
+        except errors.InvalidReferenceError:
+            self.skipTest("feature not implemented")
```
fmoor, line 69: Is there a better way to skip this test only on the stable version and not nightly? Or rather only for the server versions that don't support it?

fmoor:
I forgot to separate the flags and cardinality enum. Fixing WIP.

tests/test_datetime.py
1st1: Well, you can introspect the DB and check if the `cal::relative_duration` type is defined. If not - skip the test. We generally trust the introspection. Alternatively, you can query the server version with https://www.edgedb.com/docs/edgeql/funcops/sys#function::sys::get_version

1st1 (approved):

PR #197, opened 2021-07-07
Use system config dir instead of ~/.edgedb

fantix:
Use `edgedb self migrate` to migrate the old dir to the new location.

### DIFF
--- a/edgedb/credentials.py
+++ b/edgedb/credentials.py
@@ -1,6 +1,5 @@
 import os
 import pathlib
-import sys
 import typing
 import json
 
@@ -8,6 +7,8 @@ try:
     from typing import TypedDict
 except ImportError:
     from typing_extensions import TypedDict
+
+from . import platform
 
 
 class RequiredCredentials(TypedDict, total=True):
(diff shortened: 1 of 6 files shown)


edgedb/credentials.py
```diff
@@ -19,6 +22,10 @@ class Credentials(RequiredCredentials, total=False):
     database: typing.Optional[str]
 
 
+def get_credentials_path(instance_name: str) -> pathlib.Path:
+    return platform.config_dir() / "credentials" / (instance_name + ".json")
```
elprans, line 26: I'd put an existence check here and fallback to `~/.edgedb` to preserve compatibility with unmigrated layout.  Otherwise a binding update done before `edgedb self migrate` will break user apps.

fantix: Added in `config_dir()` so that `edgedb.toml` project link could also be compatible.

tailhook:
I'm not sure this is how fallback should work. With current impl if I create `~/.edgedb` everything breaks. I think this can be done by mistakenly using old CLI or by some other ignorant software. Probing the exact file in both locations is much more reliable.

Note: it's very different from what we do in CLI, because in CLI we actually warn and suggest you run `self migrate` that fixes the issue. For language bindings, it might run on CI, container or whatever where CLI is not available, or where warnings aren't visible.

fantix:
> With current impl if I create `~/.edgedb` everything breaks.

It's the opposite: if `~/.config/edgedb` exists, it won't fallback to `~/.edgedb`.

tailhook:
> @tailhook
> 
> > With current impl if I create `~/.edgedb` everything breaks.
> 
> It's the opposite: if `~/.config/edgedb` exists, it won't fallback to `~/.edgedb`.

Well, maybe it's better than the opposite. Anyway, I've proposed to just have kinda "search path" instead. But I would defer final decisiona

fantix:
@tailhook something like [this](https://github.com/edgedb/edgedb-python/pull/197/commits/5b272771d38db2c60b4fd4afe715b11e507734c8)?

tailhook:
> @tailhook something like [this](https://github.com/edgedb/edgedb-python/pull/197/commits/5b272771d38db2c60b4fd4afe715b11e507734c8)?

Yes, like this. Except if I'm not mistaken, it looks like if there is no respective file/dir it will return path in old config dir. But it should rather return the new one if no one exists (i.e. to print an error like `credentials file at {} does not exists` with the new path). I've untested this, so maybe I'm missing something.

tailhook (approved):

PR #198, opened 2021-07-07
TLS and ALPN support

fantix:
Depends on #197 

The test is failing due to a circular dependency of edgedb-server and edgedb-python. I'm thinking maybe we should run the server test in non-TLS mode first, and switch back to TLS when things are stable.


elprans (approved):

elprans (commented):
LGTM, but this needs some tests and a documentation update.

fantix:
Thanks!

> this needs some tests and a documentation update

Yes! I got that in my backlog. Do you think we could merge this first and fix the failing tests afterwards?

fantix:
https://github.com/edgedb/edgedb-python/pull/199/checks at least the stable server tests passed 🤞

edgedb/con_utils.py
```diff
@@ -127,10 +129,26 @@ def _stash_path(path):
     return platform.search_config_dir('projects', dir_name)
 
 
+def _parse_verify_hostname(val: str) -> typing.Optional[bool]:
+    val = val.lower()
+    if val in {"1", "yes", "true", "y", "t", "on"}:
+        return True
+    elif val in {"0", "no", "false", "n", "f", "off"}:
+        return False
+    elif val in {"default", "d"}:
```
fmoor, line 138: Why is this a possible value? If someone wants the default wouldn't they just omit the field all together? Then this could collapse to a boolean value instead of a string and there would be less parsing logic.

fantix: You're right about it. Omitting is a more reasonable option than `=default`. I'll push a fix.

PR #201, opened 2021-07-14
Remove DSN support of ?tls_verify_hostname=default

fantix:
We don't really need this value - the user should just omit the
parameter to utilize the default behavior.


fmoor (approved):
LGTM :+1:

elprans (approved):

PR #202, opened 2021-07-14
Rename --tls-cert-file to --tls-ca-file

fantix:
@fmoor if this is approved, we'd need the Go and JS bindings aligned.

Also fixed tests running on the new server, locally passed.

There will be another PR after we have the explicit option to generate dev certs on the server.

### DIFF
--- a/edgedb/con_utils.py
+++ b/edgedb/con_utils.py
@@ -145,10 +145,10 @@ def _parse_verify_hostname(val: str) -> typing.Optional[bool]:
 
 def _parse_connect_dsn_and_args(*, dsn, host, port, user,
                                 password, database, admin,
-                                tls_cert_file, tls_verify_hostname,
+                                tls_ca_file, tls_verify_hostname,
                                 connect_timeout, server_settings):
     using_credentials = False
-    tls_cert_data = None
+    tls_ca_data = None
 
     if admin:
         warnings.warn(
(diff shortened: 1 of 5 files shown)


edgedb/con_utils.py
```diff
@@ -297,8 +297,8 @@ def _parse_connect_dsn_and_args(*, dsn, host, port, user,
             password = creds['password']
         if database is None and 'database' in creds:
             database = creds['database']
-        if tls_cert_file is None and 'tls_cert_data' in creds:
-            tls_cert_data = creds['tls_cert_data']
+        if tls_ca_file is None and 'tls_cert_data' in creds:
+            tls_ca_data = creds['tls_cert_data']
```
fantix, line 301: `creds['tls_cert_data']` is from the credential JSON file, we'd have to update the Rust and CLI code if we also want to fix that.

fmoor:
> @fmoor if this is approved, we'd need the Go and JS bindings aligned.

roger :+1:

edgedb/con_utils.py
fmoor: It seems like the naming should be consistent. I think changing `cert` to `ca` everywhere is a good idea.

elprans: No. `tls_cert_data` is a valid name.  What is `cert` on the server side becomes a `ca` on the client side if you want to trust the specified cert directly.

fmoor: Aren't the `credentials/*.json` files client specific files? The server doesn't know about them.

I'm thinking that if I as a developer edit or create a `credentials/*.json` file it is easier if the names in that file line up with the names used in the bindings so that I don't have to remember two sets of names for a very similar thing.

elprans: > I don't have to remember two sets of names for a very similar thing

But they aren't the same.  `tls_cert_data` includes the server certificate specifically.  In certain cases we want to trust that cert directly and so we "pretend" that the certificate is it's own CA and so pass it in `tls_ca_file`.

elprans (approved):

PR #203, opened 2021-07-14
tests: Hide server output by default

elprans:
Point server output to `/dev/null` unless `EDGEDB_DEBUG_SERVER`
environment variable is set.


fantix (approved):

fmoor (approved):
:+1:

PR #204, opened 2021-07-15
Use new server with --generate-self-signed-cert

fantix:
Depends on edgedb/edgedb#2680

I have manually tested this locally with the server on edgedb/edgedb#2680 and the test is a pass. I haven't tested CI but I think it should work once we have the new server nightly. (We could add a generate-cert option to the setup-edgedb action later).

### DIFF
--- a/edgedb/_testbase.py
+++ b/edgedb/_testbase.py
@@ -85,6 +85,7 @@ def _start_cluster(*, cleanup_atexit=True):
             f"--emit-server-status={status_file_unix}",
             "--port=auto",
             "--auto-shutdown",
+            "--generate-self-signed-cert",
             "--bootstrap-command=ALTER ROLE edgedb { SET password := 'test' }",
         ]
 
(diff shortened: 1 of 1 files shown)


edgedb/_testbase.py
```diff
@@ -85,6 +85,7 @@ def _start_cluster(*, cleanup_atexit=True):
             f"--emit-server-status={status_file_unix}",
             "--port=auto",
             "--auto-shutdown",
+            "--generate-self-signed-cert",
```
elprans, line 88: This wouldn't work with older servers, we need to add a version check here, or react to unknown option error

fantix: Ah right, let me fix that.

fantix: Fixed - I think an option check in `--help` is simpler (maybe more reliable?) than a version check?

elprans (approved):

PR #207, opened 2021-07-16
Fix failing tests regarding --help

fantix:
`shell=True` seems to be the problem of the weird behavior on different environments. Also the `PYTHONPATH` was stopping the `--help` to import the correct uvloop. There was also a typo for Windows.

Also, the cert path in WSL is not directly translated into a path accessible from the host, so had to copy the cert to the tmpdir in the host on Windows.

### DIFF
--- a/edgedb/_testbase.py
+++ b/edgedb/_testbase.py
@@ -78,7 +78,12 @@ def _start_cluster(*, cleanup_atexit=True):
             .lower()
         )
 
-        edgedb_server = os.environ.get('EDGEDB_SERVER_BINARY', 'edgedb-server')
+        env = os.environ.copy()
+        # Make sure the PYTHONPATH of _this_ process does
+        # not interfere with the server's.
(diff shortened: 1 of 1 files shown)


edgedb/_testbase.py
```diff
@@ -91,19 +96,21 @@ def _start_cluster(*, cleanup_atexit=True):
 
         help_args = [edgedb_server, "--help"]
         if sys.platform == 'win32':
-            help_args = ['wsl', '-u', 'edgedb'] + args
+            help_args = ['wsl', '-u', 'edgedb'] + help_args
 
-        if "--generate-self-signed-cert" in subprocess.getoutput(help_args):
+        if "--generate-self-signed-cert" in subprocess.run(
```
elprans, line 101: So, on WSL you actually need to translate the cert file path back to Windows path, essentially the reverse of the status file translation above

elprans (approved):

PR #208, opened 2021-07-23
Fix confusing error when no arguments are passed to a query with named params

elprans:
Currently, if one attempts to execute a query with named parameters
without passing any arguments at all, they'll get a confusing `expected
named arguments, got positional arguments`.  Fix this to have the server
issue the appropriate error about argument counts instead.  See
edgedb/edgedb#1427.


fantix (approved):
LGTM

1st1 (approved):

PR #210, opened 2021-08-06
Fix asyncio SSL unclosed warning

fantix:
Sample test run: https://github.com/fantix/edgedb/runs/3244328000

### DIFF
--- a/edgedb/asyncio_con.py
+++ b/edgedb/asyncio_con.py
@@ -107,6 +107,7 @@ class _AsyncIOConnectionImpl:
         factory = functools.partial(
             asyncio_proto.AsyncIOProtocol, params, loop
         )
+        tr = None
 
         try:
             if isinstance(addr, str):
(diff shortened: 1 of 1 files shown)


elprans (approved):

edgedb/asyncio_con.py
```diff
@@ -155,6 +166,13 @@ async def aclose(self):
             try:
                 self._protocol.terminate()
                 await self._protocol.wait_for_disconnect()
+
+                # With asyncio on CPython 3.10 or lower, a normal exit
+                # (connection closed by peer) cannot set the transport._closed
+                # properly, leading to false ResourceWarning. Let's fix that by
+                # closing the transport again.
+                if not self._transport.is_closing():
+                    self._transport.close()
```
1st1, line 175: Should this be copied to `terminate()`?

fantix: I think `_AsyncIOConnectionImpl.terminate()` is fine, because it calls `AsyncIOProtocol.abort()` which closes the transport.

fantix: But maybe moving this logic to `AsyncIOProtocol.connection_lost()` is a better idea - so that the transport is always marked as closed as far as `connection_lost()` is called.

PR #222, opened 2021-08-14
Fix retrying_transaction() on network errors

fmoor:
Errors from blocking socket operations are wrapped with EdgeDB errors.

### DIFF
--- a/edgedb/protocol/asyncio_proto.pyx
+++ b/edgedb/protocol/asyncio_proto.pyx
@@ -49,7 +49,7 @@ cdef class AsyncIOProtocol(protocol.SansIOProtocol):
 
     cdef write(self, WriteBuffer buf):
         if self.transport is None:
-            raise ConnectionAbortedError
+            raise errors.ClientConnectionFailedTemporarilyError()
         self.transport.write(memoryview(buf))
 
     async def wait_for_message(self):
--- a/edgedb/protocol/blocking_proto.pyx
+++ b/edgedb/protocol/blocking_proto.pyx
@@ -22,7 +22,8 @@ from edgedb.pgproto.pgproto cimport (
     ReadBuffer,
 )
 
-
+from .. import con_utils
+from .. import errors
 from . cimport protocol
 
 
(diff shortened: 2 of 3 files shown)


1st1:
@fantix please review, you're working with this code right now so you definitely know its quirks better than I

edgedb/retry.py
```diff
@@ -37,11 +39,17 @@ async def __aexit__(self, extype, ex, tb):
...
+        if reconnect:
+            await self._owner._reconnect()
 
-        if (
-            extype is not None and
+        if extype is not None and (
             issubclass(extype, errors.EdgeDBError) and
             ex.has_tag(errors.SHOULD_RETRY)
+        ) or (
+            issubclass(extype, ConnectionError)
```
fantix, line 52: If I understood this correctly - we want to `_reconnect()` on all ConnectionErrors, but only retry on the ones that is not from `commit()`? A comment would be super helpful 🙏

edgedb/retry.py
```diff
@@ -37,11 +39,17 @@ async def __aexit__(self, extype, ex, tb):
             # to propagate it to cancel the whole task.
             # NOTE: rollback error is always swallowed, should we use
             # on_log_message for it?
+        except ConnectionError:
+            reconnect = True
+
+        if reconnect:
+            await self._owner._reconnect()
```
fantix, line 46: Hmm what if the `_reconnect()` fails again, shall we continue with the retry loop?

fantix (commented):
Looks good to me in general, have a question though.

edgedb/retry.py
elprans: We don't retry on network errors during `COMMIT`, because there is no way to know if it actually reached the server or not.  Also, there's no point in retrying `ROLLBACK`, since its issuance already indicated a non-retryable condition.

edgedb/retry.py
elprans: We should retry according to the retry policy.  In fact, it is likely that a reconnect fails if there are network issues that caused the original network error in the first place.

edgedb/retry.py
fmoor: changed

edgedb/retry.py
fmoor: changed. I'm not 100% I got this part right. @fantix pleas have a look :mag:

fantix (changes requested):
Per discussion, I think we should repurpose this PR to wrap all eligible ConnectionError raised from edgedb-python into one of the retryable errors:

* `ClientConnectionFailedTemporarilyError`
* `ClientConnectionTimeoutError`
* `ClientConnectionClosedError`

So that we don't even have to change the logic in `__aexit__()` of `retry.py` - the connection will try to reconnect automatically on the next acquisition and query. And for a pool we're using LIFO queue, so it's almost always reconnected on the next retry.

Please see also #230

fantix (changes requested):
Good catch on the blocking transaction retries!

* In `protocol.pyx`, there is still one usage of `ConnectionError` in `SansIOProtocol.terminate()` which should be changed.
* Some auto test would be great, but this can be a separate PR.

edgedb/retry.py
```diff
@@ -158,6 +158,11 @@ def __enter__(self):
         return self
 
     def __exit__(self, extype, ex, tb):
+        # force reconnect  on network errors
+        print(extype)
```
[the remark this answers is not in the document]
fmoor: changed

edgedb/retry.py
```diff
@@ -158,6 +158,11 @@ def __enter__(self):
...
+        if extype and issubclass(extype, errors.ClientConnectionError):
+            self._connection_inner._impl = None
```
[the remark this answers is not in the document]
fmoor: changed

fantix (approved):

PR #228, opened 2021-08-19
Retry if start() raises a retryable error

fantix:
This issue was found in HA failover when the master Postgres is down, it raises a `BackendUnavailableError` to start a transaction, but the `retrying_transaction()` couldn't capture this retryable error from `start()`, because `start()` is called in `__enter__()` and the capture happens in `__exit__()`.

~This PR captures retryable errors in the `start()` of `retrying_transaction()`, marking the transaction with an error state which will cause the first query to raise the wrapped retryable error (or the state is checked in `__exit__()` too), so that the transaction can be properly retried.~

This PR solves the problem by making the `retrying_transaction()` start lazily when there is a `tx.execute()` or `tx.query*()` call, so that `__exit__()` wll be able to capture the retryable errors.

### DIFF
--- a/edgedb/retry.py
+++ b/edgedb/retry.py
@@ -23,6 +23,9 @@ class AsyncIOIteration(_transaction.BaseAsyncIOTransaction):
             await self._start(single_connect=self.__iteration != 0)
 
     async def __aenter__(self):
+        if self._managed:
+            raise errors.InterfaceError(
+                'cannot enter context: already in an `async with` block')
         self._managed = True
         return self
 
@@ -133,6 +136,9 @@ class Iteration(_transaction.BaseBlockingIOTransaction):
             self._start(single_connect=self.__iteration != 0)
 
     def __enter__(self):
+        if self._managed:
+            raise errors.InterfaceError(
+                'cannot enter context: already in a `with` block')
         self._managed = True
         return self
 
--- a/edgedb/retry.py
+++ b/edgedb/retry.py
@@ -5,23 +5,37 @@ from . import errors
 from . import transaction as _transaction
 
 
-class AsyncIOIteration(_transaction.AsyncIOTransaction):
+class AsyncIOIteration(_transaction.BaseAsyncIOTransaction):
     def __init__(self, retry, owner, iteration):
         super().__init__(owner, retry._options.transaction_options)
         self.__retry = retry
         self.__iteration = iteration
+        self.__started = False
 
-    async def start(self):
+    async def _ensure_transaction(self):
         if not self._managed:
             raise errors.InterfaceError(
                 "Only managed retriable transactions are supported. "
                 "Use `async with transaction:`"
             )
-        await self._start(single_connect=self.__iteration != 0)
+        if not self.__started:
+            self.__started = True
+            await self._start(single_connect=self.__iteration != 0)
+
+    async def __aenter__(self):
+        self._managed = True
+        return self
 
     async def __aexit__(self, extype, ex, tb):
+        self._managed = False
+        if not self.__started:
+            return False
+
         try:
-            await super().__aexit__(extype, ex, tb)
+            if extype is not None:
+                await self._rollback()
+            else:
+                await self._commit()
         except errors.EdgeDBError as err:
             if ex is None:
                 # On commit we don't know if commit is succeeded before the
@@ -101,23 +115,37 @@ class Retry(BaseRetry):
         return iteration
 
 
-class Iteration(_transaction.Transaction):
+class Iteration(_transaction.BaseBlockingIOTransaction):
     def __init__(self, retry, owner, iteration):
         super().__init__(owner, retry._options.transaction_options)
         self.__retry = retry
         self.__iteration = iteration
+        self.__started = False
 
-    def start(self):
+    def _ensure_transaction(self):
         if not self._managed:
             raise errors.InterfaceError(
                 "Only managed retriable transactions are supported. "
                 "Use `with transaction:`"
             )
-        self._start(single_connect=self.__iteration != 0)
+        if not self.__started:
+            self.__started = True
+            self._start(single_connect=self.__iteration != 0)
+
+    def __enter__(self):
+        self._managed = True
+        return self
 
     def __exit__(self, extype, ex, tb):
+        self._managed = False
+        if not self.__started:
+            return False
+
         try:
-            super().__exit__(extype, ex, tb)
+            if extype is not None:
+                self._rollback()
+            else:
+                self._commit()
         except errors.EdgeDBError as err:
             if ex is None:
                 # On commit we don't know if commit is succeeded before the
--- a/edgedb/transaction.py
+++ b/edgedb/transaction.py
@@ -105,29 +105,8 @@ class BaseTransaction:
             mod, self.__class__.__name__, ' '.join(attrs), id(self))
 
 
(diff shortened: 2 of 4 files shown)


fantix:
Problem in this PR is that, user code will be executed if start() already failed. An alternative is to call start() in the iterator (__next__()) so that we don't have to do the BaseException hack.

edgedb/retry.py
```diff
@@ -5,21 +5,50 @@
...
 
 
+class _RetryErrorInTxBegin(BaseException):
+    # In case when it failed to start() a transaction with a retryable error,
+    # the following query will raise this error so that the transaction can be
+    # retried. This is intentionally a BaseException because we don't want user
+    # code to capture this error and end up with a broken transaction.
+
+    def __init__(self, underlying):
+        self.underlying = underlying
```
1st1, line 15:
```suggestion
    def __init__(self, wrapped: BaseExceptions):
        self.wrapped = wrapped
```

edgedb/retry.py
```diff
@@ -5,21 +5,50 @@
 from . import transaction as _transaction
 
 
+class _RetryErrorInTxBegin(BaseException):
```
1st1, line 8:
```suggestion
class _RetryableErrorInTxBegin(BaseException):
```

edgedb/retry.py
fantix: Oh the wrapped exception is a retryable `EdgeDBError` - but I'll rename, thanks!

1st1 (approved):

edgedb/retry.py
1st1: Yeah, add some docs too.

1st1:
> Problem in this PR is that, user code will be executed if start() already failed. An alternative is to call start() in the iterator (__next__()) so that we don't have to do the BaseException hack.

Actually that might be a better solution. Can you try that?

fantix:
Sure thing, one sec

edgedb/retry.py
```diff
@@ -10,16 +10,25 @@ def __init__(self, retry, owner, iteration):
...
     async def start(self):
         if not self._managed:
             raise errors.InterfaceError(
                 "Only managed retriable transactions are supported. "
                 "Use `async with transaction:`"
             )
-        await self._start(single_connect=self.__iteration != 0)
+
+    async def _ensure_transaction(self):
+        await self.start()
```
1st1, line 23: Hm, `start()` has no awaits in it now... And does pretty much nothing too.

edgedb/retry.py
```diff
@@ -10,16 +10,25 @@ def __init__(self, retry, owner, iteration):
...
+        if not self.__started:
+            self.__started = True
+            await self._start(single_connect=self.__iteration != 0)
```
1st1, line 26: Should `self.__started` be set in a `try..finally`?

1st1: Or just after the `await` line?

edgedb/retry.py
fantix: Right, it's a bit weird here - I was just trying to avoid overriding `__aenter__` here. Let me extract the `raise InterfaceError` part in `start()` to a new sync method.

1st1 (commented):
The changes look good overall, but I have more questions now:

If I explicitly call `await tx.commit(); await tx.query('SELECT 42;`)` the API would allow me that. But I think it shouldn't. Basically the new `_ensure_transaction()` should also check if the transaction was explicitly committed or rolled back, and if it was - it should error out.

edgedb/retry.py
1st1: We can keep `start()` to explicitly open a transaction. I think it's OK. (We've chatted about this in Slack)

edgedb/retry.py
fantix: Like this?

```python
    async def _ensure_transaction(self):
        await self.start()  # only managed transaction check
        if not self.__started:
            try:
                await self._start(single_connect=self.__iteration != 0)
            finally:
                self.__started = True
```

And the reason is - I'm guessing wildly - to allow running `self._start()` concurrently and one of them could raise a more reasonable error? That's actually a good point to have better error messages for wrong concurrency - but `try..finally` won't work, I'll add an `else` branch to check the transaction status and raise friendly error messages.

fantix:
```python
async for tx in con.retrying_transaction():
    async with tx:
        await tx.commit()
        await tx.execute("SELECT 42")
```

For now this will raise an error, because it's a managed transaction and `commit()` will check and raise an `InterfaceError`.

```python
async for tx in con.retrying_transaction():
    await tx.execute("SELECT 31")
    await tx.commit()
    await tx.execute("SELECT 42")
```

This doesn't work either in this PR, because `tx.start()` (called in `execute()`) will raise an `InterfaceError` if called without a managed transaction.

edgedb/retry.py
1st1: > And the reason is - I'm guessing wildly - to allow running self._start() concurrently

Ah, sure, keep it as is and disregard my earlier suggestion to move `started = True` after await.

1st1:
> For now this will raise an error, because it's a managed transaction and commit() will check and raise an InterfaceError.

Well, in this case, the "managed transaction" object that we yield from `retrying_transaction()` shouldn't even have those methods.

But then... this is a perfectly valid use case:

```python
async for tx in con.retrying_transaction():
  async with tx:
     report = await calculate_report_maybe_with_some_inserts(tx)
     await tx.rollback()
```

In the above script we do want to retry computation of the report if a network error happened, but ultimately the transaction must be reverted.

Ultimately, it seems to me that having explicit `start()`, `commit()`, and `rollback()` methods on the managed retrying transaction is simply wrong, they are not even currently working. And for the "rollback" example above perhaps we could do it with a flag to `retrying_transaction()`, like `rollback=True`:

```python
async for tx in con.retrying_transaction(rollback=True):
  async with tx:
     report = await calculate_report_maybe_with_some_inserts(tx)
```

edgedb/retry.py
elprans: Please elaborate why an explicit `start()` is necessary.

elprans (changes requested):

edgedb/retry.py
1st1: That comment was before we opened the discussion about this api. I think that we should strip all manual control menthols from the retrying transaction.

edgedb/retry.py
```diff
@@ -5,23 +5,37 @@
...
                 "Only managed retriable transactions are supported. "
                 "Use `async with transaction:`"
             )
-        await self._start(single_connect=self.__iteration != 0)
+        if not self.__started:
+            self.__started = True
+            await self._start(single_connect=self.__iteration != 0)
+
+    async def __aenter__(self):
+        self._managed = True
```
1st1, line 26: Can we add some code to prevent entering the same transaction object twice?

1st1 (approved):

edgedb/retry.py
fantix: Sure thing!

elprans (approved):

fmoor (approved):

fantix (commented):
Because of the lazy start, the `_borrowed_for` check is currently broken, allowing the following code to run:

```python
async for tx in con.retrying_transaction():
    async with tx:
        con.query("SELECT 123")  # <- con, not tx
```

I'll create a following PR to fix this.

PR #229, opened 2021-08-23
support protocol version 0.12

fmoor:
- change query argument type from tuple to object
- implement object encoding
- bump max protocol version to 0.12


edgedb/protocol/codecs/object.pyx
```diff
@@ -21,7 +21,74 @@
 cdef class ObjectCodec(BaseNamedRecordCodec):
 
     cdef encode(self, WriteBuffer buf, object obj):
```
elprans, line 23: Let's name this `encode_args` as the method is specific to argument encoding.

edgedb/protocol/codecs/object.pyx
```diff
@@ -21,7 +21,74 @@
...
+
+        for i in range(len(self.fields_codecs)):
+            name = datatypes.record_desc_pointer_name(self.descriptor, i)
+            required_args.add(name)
+
+        passed_args = set(args.keys())
+        missed_args = required_args - passed_args
+        extra_args = passed_args - required_args
+
+        error_message = f'expected {required_args} keyword arguments'
```
elprans, line 80:
```suggestion
        error_message = f'expected {required_args} arguments'
```

This isn't specific to keyword arguments, so mentioning them is confusing.  I don't think we need to be specific here, as we don't allow mixing argument styles, so simply saying "arguments" is enough.

edgedb/protocol/protocol.pyx
```diff
@@ -1104,6 +1104,17 @@ cdef class SansIOProtocol:
...
+            if type(in_dc) is not ObjectCodec:
+                raise errors.QueryArgumentError(
+                    'unexpected query argument codec')
+
+            if args:
+                kwargs = {str(i): v for i, v in enumerate(args)}
+
+            in_dc.encode(buf, kwargs)
+            return
+
```
elprans, line 1117: Indent the below block under `else: ` for clarity.

edgedb/protocol/codecs/object.pyx
fmoor: changed

edgedb/protocol/codecs/object.pyx
fmoor: changed

edgedb/protocol/protocol.pyx
fmoor: changed

elprans (approved):

PR #231, opened 2021-08-24
Fix deprecated usage of SSLContext

elprans:
`ssl.SSLContext()` without an explicit protocol argument is deprecated.


fantix (approved):
Thanks!

PR #237, opened 2021-09-01
Fix connection and retry options on AsyncIOPool

### LINKED ISSUE #185: Broken `.with_transaction_options` and `.with_retry_options` on `AsyncIOPool`
I'm on the latest edgedb-python (0.14.0) with EdgeDB 1-beta2 running in Docker

I've tried to call `with_retry_options` method on `AsyncIOPool` instance, but it fails with:

```
        ...
        result = self._shallow_clone()
>      result._options = self._options.with_retry_options(options)
E      AttributeError: 'NoneType' object has no attribute '_options'
```

I've also noticed that [AsyncIOPool](https://github.com/edgedb/edgedb-python/blob/master/edgedb/asyncio_pool.py#L544) doesn't implement `_shallow_clone` required by [`_OptionsMixin`](https://github.com/edgedb/edgedb-python/blob/master/edgedb/options.py#L117-L119).

Also, is there is any specific reason why `_OptionsMixin` doesn't inherit from `abc.ABC` but still uses `abstractmethod` decorator?

tailhook:
Fixes #185

### DIFF
--- a/edgedb/asyncio_pool.py
+++ b/edgedb/asyncio_pool.py
@@ -758,6 +758,12 @@ class AsyncIOPool(abstract.AsyncIOExecutor, options._OptionsMixin):
     def retrying_transaction(self) -> _retry.AsyncIORetry:
         return _retry.AsyncIORetry(self)
 
+    def _shallow_clone(self):
+        new_pool = self.__class__.__new__(self.__class__)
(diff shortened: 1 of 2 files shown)


fantix (approved):
LGTM

PR #243, opened 2021-09-24
Auto retry read-only queries outside transactions

fmoor:
fixes https://github.com/edgedb/edgedb-python/issues/239


edgedb/asyncio_con.py
```diff
@@ -241,12 +242,20 @@ async def ensure_connected(self, *, single_attempt=False):
...
+        try:
+            await inner._impl.connect(inner._loop, inner._addrs,
+                                      inner._config, inner._params,
+                                      single_attempt=single_attempt,
+                                      connection=self)
+        except Exception:
+            # if the new impl fails to connect it will be in an inconsistent
+            # state. Use the old impl instead because it is in a consistent
+            # state.
+            self._inner._impl = impl
```
fmoor, line 257: I'm not sure this is the best way to solve this :confused:

elprans: Call `_AsyncIOConnectionImpl.terminate()` and make sure the latter sets `_protocol` to `None`, which would make `is_closed()` return `True` and there should be no inconsistency downstream.

fmoor: changed

PR #249, opened 2021-11-03
Add optional/required `query_single*` methods + rename `retrying_transaction` + update pool to `create_client` API

jaclarke:


tests/test_async_query.py
```diff
@@ -243,15 +243,21 @@ async def test_async_query_single_command_03(self):
...
 
-        with self.assertRaisesRegex(
-                edgedb.InterfaceError,
-                r'cannot be executed with query_single_json\(\).*'
-                r'not return'):
-            await self.con.query_single_json('START TRANSACTION')
+        # self.assertEqual(
+        #     await self.con.query_single('START TRANSACTION'), None)
+
+        # with self.assertRaisesRegex(
```
elprans, line 254: uncomment?

jaclarke: I left this bit commented out for now, because of this bug: https://github.com/edgedb/edgedb/issues/3120

elprans: Let's reference this but in the comment then.

elprans (approved):

1st1:
The `connect()` function should be refactored into this (in pseudo-code):

```python
async def connect():
    warnings.warn('deprecated')
    client = create_client(concurrency=1)
    await client.ensure_connected()
    return client
```

Let's keep raw (unwrapped) connection around and add a top-level function `async connect_raw()` to return it. But let's not document it just yet.

1st1 (approved):

1st1:
@jaclarke James, everything looks great. Can you fix the CI issue and merge?

This can go in a separate PR https://github.com/edgedb/edgedb-python/pull/249#issuecomment-964627687

edgedb/asyncio_pool.py
```diff
@@ -579,33 +528,38 @@ def __init__(self, *connect_args,
...
         if self._impl._queue is None:
             # Queue has not been initialized yet
-            return self._impl._maxsize
+            return self._impl._concurrency
 
         return self._impl._queue.qsize()
 
+    async def ensure_connected(self):
+        for ch in self._impl._holders:
+            if ch._con is not None and ch._con.is_closed():
```
fantix, line 548: Just to be sure - this is not intentionally `and ch._con.is_closed()` here for some reason, right? Because I think it's a typo, it should be `and not ch._con.is_closed()`.

elprans: Yea, it's a typo

PR #252, opened 2021-11-06
Fix `credentials_file` argument typo on `async_connect`

mkniewallner:
Fixing a small typo on `async_connect` method.
As this argument was recently introduced in https://github.com/edgedb/edgedb-python/pull/241 and is not in the latest released version (`0.17.3`), it should not be a breaking change.


jaclarke (approved):

jaclarke:
Thanks @mkniewallner

PR #264, opened 2021-11-14
change type for tls_security from bool to str

nsidnev:


elprans (approved):

edgedb/asyncio_con.py
```diff
@@ -578,7 +578,7 @@ async def async_connect_raw(dsn: str = None, *,
                             user: str = None, password: str = None,
                             database: str = None,
                             tls_ca_file: str = None,
-                            tls_security: bool = None,
+                            tls_security: str = None,
```
1st1, line 581: @jaclarke should the default be `'default'` or `None`?

nsidnev: This is just IMO, but I think `None` is appropriate here. All default values for connection parameters (such as tls_security, hostname, database, etc.) are defined in the properties of `edgedb.con_utils.ResolvedConnectConfig` when they are accessed

jaclarke: Yes, it should be `None`. `'default'` would override any value given in the `EDGEDB_CLIENT_TLS_SECURITY` env var, or the dsn and credentials_file options (and their env vars).

PR #267, opened 2021-11-16
Improve IPv6 parsing

1st1:
Complimentary to https://github.com/edgedb/shared-client-testcases/pull/11

### DIFF
--- a/edgedb/con_utils.py
+++ b/edgedb/con_utils.py
@@ -284,6 +284,14 @@ def _validate_host(host):
     return host
 
 
+def _prepare_host_for_dsn(host):
+    host = _validate_host(host)
+    if ':' in host:
+        # IPv6
+        host = f'[{host}]'
+    return host
+
+
 def _validate_port(port):
     try:
         if isinstance(port, str):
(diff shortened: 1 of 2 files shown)


edgedb/con_utils.py
```diff
@@ -630,7 +638,8 @@ def _resolve_config_options(
                 resolved_config.set_port(*port)
             if dsn is None:
                 dsn = (
-                    'edgedb://' + (_validate_host(host[0]) if host else ''),
+                    'edgedb://' +
+                    (_prepare_host_for_dsn(host[0]) if host else ''),
```
elprans, line 642: Are we generating a DSN here only to parse it back again?  Is there a better way?

1st1: @jaclarke

1st1: Yeah, I'm also wondering why we're doing this that way. Also I'm a bit lost in `host[0]`, `port[0]`, etc

jaclarke: I think it was so the hostname would be validated the same (by `urllib.parse.urlparse`) whether it was given as `host` or as part of `dsn`

jaclarke: > Also I'm a bit lost in host[0], port[0], etc

The connection options are passed around as tuples of `(value, source)` to keep track of where they came from for this:

> we'll be including the source of each connection option in the connection error message if one occurs.
https://github.com/edgedb/edgedb/discussions/2922#discussioncomment-1325493

(though it seems that hasn't been fully implemented)

1st1: @elprans tests pass, I think we should merge this.

elprans (approved):

PR #274, opened 2021-11-27
safely unpack cached query data for readonly queries retries

nsidnev:
this fixes an unpacking error, when the EdgeDB server is restarted before the query is parsed, which causes `ClientConnectionError` to be raised and failure to cache query data


nsidnev:
Hmm. It looks like CI can't successfully install the EdgeDB server, although I'm not sure why CI on master is green... 

Locally all the tests are successful.

elprans:
>  It looks like CI can't successfully install the EdgeDB server

Looks like a flake, possibly a network issue.

elprans (approved):

1st1 (approved):

PR #289, opened 2022-01-21
[PoC] Timeout Implementation

fantix:
TODO: 

- [ ] Use edgedb/edgedb#3358
- [ ] Fix connect timeout in blocking client
- [ ] More tests for query timeouts using options
- [ ] More tests for query timeouts using asyncio.wait_for


edgedb/options.py
```diff
@@ -137,35 +154,54 @@ def with_transaction_options(self, options: TransactionOptions = None):
...
-        them transaction options applied will be different.
+        them retry options applied will be different.
         """
 
         result = self._shallow_clone()
         result._options = self._options.with_retry_options(options)
         return result
 
+    def with_timeout_options(self, options: TimeoutOptions = None):
+        """Returns object with adjusted options for future timeouts.
```
1st1, line 174: @tailhook @elprans do we want to use TimeoutOptions here instead of kwargs?

1st1: The api feels unnecessarily verbose to me.

1st1 (approved):
ltgm, there's an open question about the API (which can be addressed in a follow up PR). I think this is good to land as is.

edgedb/options.py
elprans: We have discussed this in edgedb/edgedb#3352 and you seemingly agreed with the mixin approach.

1st1: Hm, this is my comment: https://github.com/edgedb/edgedb/discussions/3352#discussioncomment-2036168, that specifically suggests the `client.with_timeout(command_timeout=5)` api

elprans: Ah, OK. I got confused.

We need to think how timeouts integrate with retries and if there is a benefit to a pluggable timeout policy like we have with retries.

tailhook: I think the major issue with retries is that they are more complex that one or two numbers: this is a set of rules and we had to have both, the way to add a single rule and the way to replace all rules reliably.

This is not the case for timeouts. So `with_timeout` or similar with kwargs is fine.

We can also have `with_options(command_timeout=5)`, to get rid of repeating `timeout`. However, we may want to keep `with_options` for something that is a combination of retry+transactions+timeout options, if we ever need one.

fantix:
Per discussion, we decided not to rush for timeout implementation.

PR #286, opened 2022-01-05
Implementation refactor for Client-based interface

fantix:
Review note: there is one merge commit "Partially revert ..." in this branch to reserve git history; use [this link](https://github.com/edgedb/edgedb-python/pull/286/files/5d4a0a5bd2e3ec16687fe075b189c44942983e14..1f51604e646e07eed343526ed8acaa17593362ae) to see the merged diff.

- [x] Merge Connection with RawConnection
- [x] Fix tests to use client
- [x] Add previously-deleted tests
- [ ] Complete timeout implementation and add more timeout tests (will be done in a new PR)
- [x] Drop `edgedb.connect()` and replace with a single-connection `Client`
- [x] Extract base class for Client, and further remove redundant code
- [x] Extract redundant code from transactions and only use `_iter_coroutine()` on top level
- [x] Review how pool is implemented, simplify if possible

edgedb-python-pool drawio

PR #297, opened 2022-02-18
Fix docs to say `max_concurrency` instead of `concurrency`

msullivan:
Change was made in #286


fantix (approved):
Thanks!

1st1:
This one seems to be outdated

PR #299, opened 2022-02-19
Retry on TransactionError when attempting to commit

msullivan:
Serialization errors can be generated at commit time and should be
retried.

We probably want to check the other bindings for this also?


elprans (approved):

msullivan:
I'm able to reproduce the bug manually and verify the fix works, but my attempts to reproduce it in the test suite haven't worked! Argh! Going to just merge it for now.

PR #310, opened 2022-05-17
Accept args in execute() and use the new Execute message

fantix:


edgedb/protocol/protocol.pxd
```diff
@@ -40,10 +40,11 @@ include "./codecs/codecs.pxd"
 ctypedef object (*decode_row_method)(BaseCodec, FRBuffer *buf)
 
 
-cpdef enum IoFormat:
+cpdef enum OutputFormat:
     BINARY = b'b'
     JSON = b'j'
     JSON_ELEMENTS = b'J'
+    NULL_ = b'n'
```
elprans, line 47: Not a fan of trailing underscore here.  Let's use `NONE` to avoid reserved keyword use.

elprans (approved):

PR #311, opened 2022-05-22
improve error when passing empty query arguments

nsidnev:
Quite a small change to show `nothing` instead of python's `set()` when query requires something but user hasn't provided anything.

Also protocol `0.12` has been stable for a while. Maybe driver should get rid of arguments encoding with named tuple codec + remove that part from tests?

PR #314, opened 2022-05-27
drop legacy arguments encoding with named tuple codec

nsidnev:
Continues https://github.com/edgedb/edgedb-python/pull/311

Though I haven't received an answer as to whether it's time to drop the legacy encoding, so if it's too early, feel free to close this PR.


edgedb/protocol/protocol.pyx
```diff
@@ -1089,22 +1089,15 @@ cdef class SansIOProtocol:
 
         in_dc_type = type(in_dc)
 
-        if in_dc_type in {NullCodec, EmptyTupleCodec}:
-            # TODO: drop EmptyTupleCodec when 1.0 RC1 is released.
```
nsidnev, line 1093: I am not quite sure if that means removing the codec only from this condition or removing the codec completely

nsidnev:
@fantix friendly ping :)

edgedb/protocol/protocol.pyx
fantix: We can address dropping `EmptyTupleCodec` itself in a separate PR - it shouldn't be an issue keeping it I think.

fantix (approved):
Thanks for the PR and sorry for the delay! There was some discussion and it eventually came to #318.

This PR looks good! I'll rebase #318 on top of this.

PR #317, opened 2022-06-13
Change deprecated CONFIGURE SYSTEM to CONFIGURE INSTANCE

fmoor:


elprans (approved):

edgedb/_testbase.py
```diff
@@ -393,7 +393,7 @@ def setUp(self):
         if self.INTERNAL_TESTMODE:
             self.loop.run_until_complete(
                 self.client.execute(
-                    'CONFIGURE SESSION SET __internal_testmode := true;'))
+                    'CONFIGURE INSTANCE SET __internal_testmode := true;'))
```
tailhook, line 396: It wasn't `SYSTEM` it's `SESSION`. Are you sure this has to be changed?

elprans: Oh, whoops. This change is, indeed, incorrect

PR #318, opened 2022-06-17
Stop using Parse, replace use of headers with fields

elprans:
This makes these changes to the 1.0 protocol:

1. We no longer use Parse and instead rely on Execute semantics
2. We no longer rely on message headers to pass material information
   that affects how the message is processed.  These are now proper
   fields on messages.
3. We use a known-invalid typedesc id to force Parse behavior from
   Execute.

### DIFF
--- a/edgedb/protocol/protocol.pyx
+++ b/edgedb/protocol/protocol.pyx
@@ -228,83 +228,6 @@ cdef class SansIOProtocol:
         else:
             buf.write_int16(0)  # no headers
 
-    async def _parse(
-        self,
-        query: str,
-        *,
-        reg: CodecsRegistry,
-        output_format: OutputFormat=OutputFormat.BINARY,
-        expect_one: bint=False,
-        required_one: bool=False,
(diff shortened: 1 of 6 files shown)


edgedb/protocol/protocol.pyx
```diff
@@ -1213,10 +1079,10 @@ cdef class SansIOProtocol:
 
     cdef parse_command_complete_message(self):
         assert self.buffer.get_message_type() == COMMAND_COMPLETE_MSG
-        attrs = self.parse_headers()
+        self.parse_headers()
```
1st1, line 1082: Should probably rename to ignore_headers

1st1 (approved):

fantix (approved):

edgedb/protocol/protocol.pyx
fantix: I'll re-implement `parse_headers()` in the new way with argument `ignore: bool`.

PR #319, opened 2022-06-17
Change headers format to str:json

fantix:
Headers now has the following format:

```c
struct Header {
  // Header key.
  bytes           key;
  // Header data.
  bytes           value;
}

struct SomeMessage {
  ...
  // A set of message headers.
  uint16          num_headers;
  Header          headers[num_headers];
}
```

* Header in ErrorMessage is not changed
* ClientHandshake always send b'\0\0' (empty ext headers)
* ServerHandshake no longer has ext headers


setup.py
```diff
@@ -71,7 +71,7 @@
 }
 
 
-CFLAGS = ['-O2']
+CFLAGS = ['-O0']
```
elprans, line 74: You probably didn't mean to commit this.

elprans (approved):

PR #322, opened 2022-06-22
Support EDGEDB_WAIT_UNTIL_AVAILABLE environment variable

fmoor:
fixes https://github.com/edgedb/edgedb-python/issues/302

### DIFF
--- a/edgedb/con_utils.py
+++ b/edgedb/con_utils.py
@@ -47,6 +47,28 @@ TEMPORARY_ERROR_CODES = frozenset({
     errno.ENOENT,
 })
 
+ISO_DURATION_RE = re.compile(
+    r'^PT' +
+    r'(?:(?P<hours>\d*\.?\d*)H)?' +
+    r'(?:(?P<minutes>\d*\.?\d*)M)?' +
+    r'(?:(?P<seconds>\d*\.?\d*)S)?$'
+)
+HUMAN_DURATION_HOUR_RE = re.compile(
+    r'(?P<number>\d+|\d+\.\d+|\.\d+)\s*(?:h|hours?)(?P<tail>\s|$|\d)'
+)
+HUMAN_DURATION_MINTUE_RE = re.compile(
+    r'(?P<number>\d+|\d+\.\d+|\.\d+)\s*(?:m|minutes?)(?P<tail>\s|$|\d)'
+)
+HUMAN_DURATION_SECOND_RE = re.compile(
+    r'(?P<number>\d+|\d+\.\d+|\.\d+)\s*(?:s|seconds?)(?P<tail>\s|$|\d)'
+)
+HUMAN_DURATION_MS_RE = re.compile(
+    r'(?P<number>\d+|\d+\.\d+|\.\d+)\s*(?:ms|milliseconds?)(?P<tail>\s|$|\d)'
+)
+HUMAN_DURATION_US_RE = re.compile(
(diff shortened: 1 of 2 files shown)


edgedb/con_utils.py
```diff
@@ -198,7 +222,22 @@ def set_tls_security(self, security, source):
...
+        self._set_param(
+            'wait_until_available',
+            wait_until_available,
+            source,
+            _validate_wait_until_available,
+        )
+
     def add_server_settings(self, server_settings):
+        if 'wait_until_available' in server_settings:
+            import traceback
```
elprans, line 235: Debug code leftovers

edgedb/con_utils.py
```diff
@@ -326,6 +369,76 @@ def _validate_user(user):
...
+        HUMAN_DURATION_MINTUE_RE, string)
+    second, string = _parse_human_duration_unit(
+        HUMAN_DURATION_SECOND_RE, string)
+    ms, string = _parse_human_duration_unit(HUMAN_DURATION_MS_RE, string)
+    us, string = _parse_human_duration_unit(HUMAN_DURATION_US_RE, string)
+
+    if string.strip() != '':
+        raise ValueError(f'invalid duration {string!r}')
+
+    no_value = (
```
elprans, line 407: This condition seems too forgiving to me.  It should be an error if there is _any_ unrecognized text in the passed value to avoid confusion and bugs.

elprans: Nevermind, I see the extra check.

edgedb/con_utils.py
```diff
@@ -326,6 +369,76 @@ def _validate_user(user):
...
+    )
+
+
+def _parse_human_duration_unit(re, string):
+    number = None
+    match = re.search(string)
+    if match:
+        number = match.group('number')
+        tail = match.group('tail')
+        string = re.sub(tail, string, count=1)
```
elprans, line 390: What is this `re.sub` for?

edgedb/con_utils.py
```diff
@@ -326,6 +369,76 @@ def _validate_user(user):
...
+    hour, string = _parse_human_duration_unit(HUMAN_DURATION_HOUR_RE, string)
+    minute, string = _parse_human_duration_unit(
+        HUMAN_DURATION_MINTUE_RE, string)
+    second, string = _parse_human_duration_unit(
+        HUMAN_DURATION_SECOND_RE, string)
+    ms, string = _parse_human_duration_unit(HUMAN_DURATION_MS_RE, string)
+    us, string = _parse_human_duration_unit(HUMAN_DURATION_US_RE, string)
+
+    if string.strip() != '':
+        raise ValueError(f'invalid duration {string!r}')
```
elprans, line 405: This error should use the original passed string, otherwise `1hour 1hour` results in `"invalid duration '1hour'"` which is confusing.

edgedb/con_utils.py
fmoor: Repeated units are not valid. 

```
edgedb> select <duration>'1hour 1hour';
edgedb error: InvalidValueError: invalid input syntax for type std::duration: "1hour 1hour"
```

And this enables checking for extra text in the string that is not a unit i.e. `1 hour ago we went to the store...`.

fmoor: changed to `str.replace()`

elprans (approved):

PR #323, opened 2022-06-22
Bring back Parse message

elprans:
Upon further consideration it was decided that an explicit "parse only"
mode deserves a dedicated message rather than a magical type descriptor
id or a boolean flag in `Execute`.

### DIFF
--- a/edgedb/protocol/protocol.pyx
+++ b/edgedb/protocol/protocol.pyx
@@ -202,6 +202,118 @@ cdef class SansIOProtocol:
             raise errors.ClientConnectionClosedError(
                 'the connection has been closed')
 
+    cdef WriteBuffer encode_parse_params(
+        self,
+        str query,
+        object output_format,
+        bint expect_one,
+        bint required_one,
+        int implicit_limit,
+        bint inline_typenames,
+        bint inline_typeids,
+        uint64_t allow_capabilities,
+    ):
+        cdef:
+            WriteBuffer buf
+
+        compilation_flags = enums.CompilationFlag.INJECT_OUTPUT_OBJECT_IDS
+        if inline_typenames:
+            compilation_flags |= enums.CompilationFlag.INJECT_OUTPUT_TYPE_NAMES
+        if inline_typeids:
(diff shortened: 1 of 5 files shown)


edgedb/protocol/protocol.pyx
```diff
@@ -281,7 +404,9 @@ cdef class SansIOProtocol:
                         expect_one,
                         new_cardinality == CARDINALITY_NOT_APPLICABLE,
                         in_dc, out_dc, capabilities)
-                    re_exec = True
+
+                    if in_tid != in_dc.get_tid():
```
fantix, line 408: I thought the server will raise a `ParameterTypeMismatchError` if the input type ids mismatch?

elprans: Ah, this is dead code left over from an earlier iteration.

edgedb/protocol/protocol.pyx
```diff
@@ -219,30 +331,38 @@ cdef class SansIOProtocol:
...
+            output_format=output_format,
+            expect_one=expect_one,
+            required_one=required_one,
+            implicit_limit=implicit_limit,
+            inline_typenames=inline_typenames,
+            inline_typeids=inline_typeids,
+            allow_capabilities=allow_capabilities,
+        )
+
+        if isinstance(in_dc, InvalidCodec):
```
fantix, line 355: We should probably drop `InvalidCodec`.

edgedb/protocol/protocol.pyx
```diff
@@ -219,30 +331,38 @@ cdef class SansIOProtocol:
...
+            if not args and not kwargs:
+                in_dc = NULL_CODEC
```
fantix, line 357: This `NULL_CODEC` guess could happen earlier in `def execute()` before calling `self._parse()`?

edgedb/protocol/protocol.pyx
```diff
@@ -219,30 +331,38 @@ cdef class SansIOProtocol:
...
+            else:
+                need_parse = True
```
fantix, line 359: `need_parse` is never used now.

fantix (approved):
Overall I think this is good! The null input guess could be optimized though.

Issue #296, opened 2022-02-16
Make `edgedb.Tuple` compatible with `tuple`


colinhacks:
Per https://discord.com/channels/841451783728529451/849374935209607199/943568048504328242

fmoor:
It looks like we are missing at least the following:

```
t.__add__(other: edgedb.Tuple) -> edgedb.Tuple
t.__mul__(other: int) -> edgedb.Tuple
t.__rmul__(other: int) -> edgedb.Tuple
t.__getitem__(key: slice) -> any
t.index(value: any, start: int = 0, stop: int = 9223372036854775807, /) -> int
t.count(value: any, /) -> int
```

I'm not sure if these are a good idea or not... Maybe not since it isn't clear what the return type should be.
```
t.__add__(other: tuple) -> ?
t.__radd__(other: tuple) -> ?
```

fantix:
Since 1.0 we started to use the built-in `tuple` type in Python.

PR #325, opened 2022-06-24
Support slicing edgedb.Tuple

fmoor:
Working towards https://github.com/edgedb/edgedb-python/issues/296

### DIFF
--- a/edgedb/datatypes/tuple.c
+++ b/edgedb/datatypes/tuple.c
@@ -172,6 +172,98 @@ tuple_getitem(EdgeTupleObject *o, Py_ssize_t i)
     return el;
 }
 
+static PyObject *
+tuple_getslice(EdgeTupleObject *o, Py_ssize_t start, Py_ssize_t stop, Py_ssize_t step) {
+    Py_ssize_t size = Py_SIZE(o);
+
+    // negative indexes count from the end
+    if (start < 0)
+        start += size;
+    if (stop < 0)
+        stop += size;
+
+    if (step < 0) {
+        if (stop < -1)
+            stop = -1;
+        if (start > size - 1)
+            start = size - 1;
+
+        if (start <= stop)
+            return EdgeTuple_New(0);
+    } else {
+        if (start < 0)
+            start = 0;
+        if (stop > size)
+            stop = size;
+
+        if (stop <= start)
+            return EdgeTuple_New(0);
+
+        if (start == 0 && stop == size && step == 1) {
+            Py_INCREF(o);
+            return o;
+        }
+    }
+
+    Py_ssize_t new_size = 0;
+    if (step < 0) {
+        for (Py_ssize_t i = start; i > stop; i += step) {
+            new_size += 1;
+        }
+    } else {
+        for (Py_ssize_t i = start; i < stop; i += step) {
+            new_size += 1;
+        }
+    }
+
+    PyObject *n = EdgeTuple_New(new_size);
+    Py_ssize_t j = 0;
+    if (step < 0) {
+        for (Py_ssize_t i = start; i > stop; i += step) {
+            PyObject *el = PyTuple_GET_ITEM(o, i + 1);
+            Py_INCREF(el);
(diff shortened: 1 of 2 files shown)


edgedb/datatypes/tuple.c
```diff
@@ -172,6 +172,98 @@ tuple_getitem(EdgeTupleObject *o, Py_ssize_t i)
...
+        for (Py_ssize_t i = start; i < stop; i += step) {
+            new_size += 1;
+        }
+    }
+
+    PyObject *n = EdgeTuple_New(new_size);
+    Py_ssize_t j = 0;
+    if (step < 0) {
+        for (Py_ssize_t i = start; i > stop; i += step) {
+            PyObject *el = PyTuple_GET_ITEM(o, i + 1);
```
fmoor, line 223: Why is this `i + 1` and not `i`?

fmoor:
I have never touched a c extension before :/

edgedb/datatypes/tuple.c
```diff
@@ -172,6 +172,98 @@ tuple_getitem(EdgeTupleObject *o, Py_ssize_t i)
...
+
+        if (start == 0 && stop == size && step == 1) {
+            Py_INCREF(o);
+            return o;
+        }
+    }
+
+    Py_ssize_t new_size = 0;
+    if (step < 0) {
+        for (Py_ssize_t i = start; i > stop; i += step) {
```
fmoor, line 210: I'm pretty sure there is a better way than this to calculate the new tuple's size...

elprans: See how Python does this: https://github.com/python/cpython/blob/0a40025b803a8dc1d604538e317992827ab96625/Objects/tupleobject.c#L779

edgedb/datatypes/tuple.c
```diff
@@ -172,6 +172,60 @@ tuple_getitem(EdgeTupleObject *o, Py_ssize_t i)
...
+    }
+
+    if (start == 0 && step == 1 && slicelength == size) {
+        Py_INCREF(o);
+        return o;
+    }
+
+    PyObject *result = EdgeTuple_New(slicelength);
+    for (cur = start, i = 0; i < slicelength; cur += step, i++) {
+        PyObject *el = PySequence_GetItem(o, cur);
```
elprans, line 192:
```suggestion
        PyObject *el = EdgeTuple_GET_ITEM(o, i);
```

edgedb/datatypes/tuple.c
```diff
@@ -172,6 +172,60 @@ tuple_getitem(EdgeTupleObject *o, Py_ssize_t i)
...
+        return tuple_getslice(o, start, stop, step);
+    }
+
+    if (PyIndex_Check(key)) {
+        Py_ssize_t val = PyNumber_AsSsize_t(key, PyExc_IndexError);
+        if (val == -1 && PyErr_Occurred()) {
+            return NULL;
+        }
+
+        return PySequence_GetItem(o, val);
```
elprans, line 219:
```suggestion
        return EdgeTuple_GET_ITEM(o, val);
```

elprans, line 219:
```suggestion
        Py_ssize_t val = PyNumber_AsSsize_t(key, PyExc_IndexError);
        if (val == -1 && PyErr_Occurred()) {
            return NULL;
        }

        return tuple_getitem(o, val);
```

1st1 (approved):

fmoor:
> We'll still have to have `edgedb.NamedTuple` fwiw, but I don't see any support of new slicing capabilities added to it. Why?

I had never worked on a C extension before so I was starting small.

1st1:
> I had never worked on a C extension before so I was starting small.

NP, but since this is now done and looks good, I suggest to generalize the functions into `_EdgeGeneric_*` family and reuse them from both the namedtuple and tuple (because both types support slicing etc).

1st1:
Closing this one -- we got rid of edgedb.Tuple

PR #327, opened 2022-06-28
Add state back to the Parse message

fantix:


fantix:
> What is the context here? Does compilation depend on globals? Or is it because of other session state?

Yeah, the session state may contain module aliases or config that affect compilation. I was only thinking of globals too until James told me the fact I forgot 🤦‍♂️

elprans (approved):

PR #329, opened 2022-06-30
Add StateDataDescription message

fantix:
* Also drop the state_description ParameterStatus

- [ ] Handle state desc in transaction (will address in a separate PR)


edgedb/protocol/protocol.pyx
```diff
@@ -480,14 +476,41 @@ cdef class SansIOProtocol:
...
+            WriteBuffer buf
 
         self.ensure_connected()
         self.reset_status()
 
+        if state is not None:
+            if self.state_codec is None:
+                try:
+                    await self._parse(
+                        "select 0",
```
elprans, line 493: Rather than forcing the clients to do hacks like this, why can't we push the initial `StateDataDescription` message on connection?

edgedb/protocol/protocol.pyx
```diff
@@ -150,10 +150,9 @@ cdef class SansIOProtocol:
         self.reset_status()
         self.protocol_version = (PROTO_VER_MAJOR, 0)
 
-        self.state_type_id = NULL_CODEC_ID
+        self.state_type_id = INVALID_CODEC_ID
```
elprans, line 153: I don't think we need to do this anymore, do we?

fantix: Right, good catch!

elprans (approved):

edgedb/protocol/protocol.pyx
```diff
@@ -249,12 +234,10 @@ cdef class SansIOProtocol:
         buf.write_byte(output_format)
         buf.write_byte(CARDINALITY_ONE if expect_one else CARDINALITY_MANY)
         buf.write_len_prefixed_utf8(query)
-        if self.state is not None:
-            buf.write_bytes(self.state_type_id)
-            buf.write_buffer(self.state)
-        else:
-            buf.write_bytes(NULL_CODEC_ID)
```
elprans, line 256: I'm not sure this change is sound.  If the client has no state to send, then sending `NULL` is preferable, because then this reduces the risk of concurrent DDL crashing queries, which affects our test suite, see https://github.com/edgedb/edgedb/runs/7141764907?check_suite_focus=true#step:17:605

elprans: Hm, nevermind `encode_state` seems to do this.

PR #335, opened 2022-07-07
Implement support for cal::date_duration

elprans:


edgedb/datatypes/date_duration.pyx
```diff
@@ -0,0 +1,87 @@
...
+    def __str__(self):
+        # This is closely modeled after interval_out().
+        cdef:
+            int64_t year = self.months / MONTHS_PER_YEAR
+            int64_t mon = self.months % MONTHS_PER_YEAR
+            int64_t tfrac
+            int64_t hour
+            int64_t min
+            int64_t sec
+            int64_t fsec
```
vpetrovykh, line 58: Do we need these for DateDuration?

edgedb/protocol/codecs/codecs.pyx
```diff
@@ -451,6 +451,35 @@ cdef relative_duration_decode(pgproto.CodecContext settings, FRBuffer *buf):
...
+    buf.write_int64(0)
+    buf.write_int32(days)
+    buf.write_int32(months)
+
+
+cdef date_duration_decode(pgproto.CodecContext settings, FRBuffer *buf):
+    cdef:
+        int32_t days
+        int32_t months
+        int64_t microseconds
```
vpetrovykh, line 471: Why are there microseconds here? Shouldn't the server reliably send the correct data type, the one without fractional day parts?

elprans: See the assertion below.

edgedb/datatypes/date_duration.pyx
elprans: Right, these are unused.

edgedb/protocol/codecs/codecs.pyx
elprans: To clarify, there is no "data type without fractions" on the wire, we use `interval` for all durations.

vpetrovykh (approved):

1st1 (approved):

msullivan (approved):

PR #340, opened 2022-07-12
Encode `None` as an explicit empty set for in `with_globals`

msullivan:
Optional global variables with a default need to be able to
distinguish between the "unset" case (where the default is used) and
the case where the global is *explicitly* set to `{}` (in which case
the default is *not* used).

Add some simple tests for globals, too.


elprans (approved):

tailhook:
This has to be documented here: https://www.edgedb.com/docs/reference/protocol/dataformats#sparse-objects

PR #347, opened 2022-07-18
Recover from failed COMMIT

fantix:
If `COMMIT` fails, run `ROLLBACK` to save the connection.

### DIFF
--- a/edgedb/transaction.py
+++ b/edgedb/transaction.py
@@ -139,6 +139,9 @@ class BaseTransaction:
                 await self._privileged_execute(query)
             except BaseException:
                 self._state = TransactionState.FAILED
+                if extype is None:
+                    # COMMIT itself may fail; recover in connection
+                    await self._privileged_execute("ROLLBACK;")
                 raise
             else:
                 self._state = state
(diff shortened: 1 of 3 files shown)


elprans (approved):

edgedb/transaction.py
```diff
@@ -139,6 +139,9 @@ async def _exit(self, extype, ex):
                 await self._privileged_execute(query)
             except BaseException:
                 self._state = TransactionState.FAILED
+                if extype is None:
+                    # COMMIT itself may fail; recover in connection
+                    await self._privileged_execute("ROLLBACK;")
```
tailhook, line 144: I think we need `self._state = TransactionState.ROLLEDBACK` ? Although, I'm not sure what the state is for.

fantix: Looks like `self._state` was used to rule the error messages when transaction control APIs are called at the wrong time. It's a bit unlikely to happen now but I'd keep it for the moment. `FAILED` seems right because the commit did fail; would be confusing to see `the transaction is already rolled back` after trying to commit.

PR #354, opened 2022-07-25
Improve duration parsing

fmoor:
The initial EDGEDB_WAIT_UNTIL_AVAILABLE duration parsing had a lot of holes in it. This fix makes it much more accurate.


vpetrovykh (approved):
I guess I generally trust the tests themselves. So that's a good indication that it works.

Depending on how complete or similar to allowed `<duration>` values you want to get you should also include the following "human" units:
1) hours - `hr`, `hrs`
2) minutes - `min`, `mins`
3) seconds - `sec`, `secs`
4) milliseconds - `msec`, `msecs`
5) microseconds - `usec`, `usecs`

I think the tests don't include "1 millisecond" and "1 microsecond" either, although I can see them in the regexps. And maybe the uppercase versions of units should also be tested since the regexp seems to indicate that it's case-insensitive.

fmoor:
> you should also include the following "human" units

I'll add this to my todo list.

Issue #127, opened 2020-11-03
Object's __hash__ and __eq__ are out of sync


1st1:
`Object.__hash__` takes into account all fetched pointers, essentially the hashing is similar to tuples.

`Object.__eq__` and other compare operators compare by object's EdgeDB ID.

fmoor:
If I remember correctly the plan is to no longer use implicit `id`s in this driver so `Object.__eq__` should probably change to mimic tuples. Does this sound right?

PR #355, opened 2022-07-27
Sync Object __hash__ and __eq__

fmoor:
Fixes https://github.com/edgedb/edgedb-python/issues/127


1st1 (commented):
What are we using for `__hash__` now?

I'm not sure we should compare based on contents but instead just inherit the standard `object` comparison ops.

fmoor:
> What are we using for `__hash__` now?

currently we use `_EdgeGeneric_HashWithBase()`

1st1:
@fantix this seems to be outdated?

PR #359, opened 2022-08-15
Implement dataclass for EdgeObject

fantix:
This would allow using the `EdgeObject` as a dataclass, in particular:

* `dataclasses.is_dataclass(obj)` will return `True`
* `dataclasses.asdict(obj)` will return a proper `dict`
* No impact to performance

Then we could do something like this with FastAPI as an example:

```python
import fastapi
import edgedb
import uuid
from dataclasses import dataclass

app = fastapi.FastAPI()
client = edgedb.create_client()

@dataclass
class Role:
    id: uuid.UUID
    name: str

    @classmethod
    def __get_validators__(cls):
        return []

@dataclass
class User:
    id: uuid.UUID
    name: str
    role: Role

    @classmethod
    def __get_validators__(cls):
        return []

@app.get("/", response_model=User)
def read_root():
    return client.query("select User { name, role: { name } }")[0]
```

Schema:

```
module default {
    type User {
        property name -> str;
        link role -> Role;
    }

    type Role {
        required property name -> str;
    }
}
```

### DIFF
--- a/edgedb/datatypes/object.c
+++ b/edgedb/datatypes/object.c
@@ -189,6 +189,10 @@ object_hash(EdgeObject *o)
 static PyObject *
 object_getattr(EdgeObject *o, PyObject *name)
 {
+    // Used in `dataclasses.as_dict()`
+    if (PyUnicode_CompareWithASCIIString(name, "__dataclass_fields__") == 0) {
+        return EdgeRecordDesc_GetDataclassFields((PyObject *)o->desc);
+    }
     Py_ssize_t pos;
     edge_attr_lookup_t ret = EdgeRecordDesc_Lookup(
         (PyObject *)o->desc, name, &pos);
@@ -364,6 +368,15 @@ EdgeObject_InitType(void)
     if (PyType_Ready(&EdgeObject_Type) < 0) {
         return NULL;
     }
+
+    // Pass the `dataclasses.is_dataclass(obj)` check, the dict is always empty
+    PyObject *default_fields = PyDict_New();
+    if (default_fields == NULL) {
+        return NULL;
+    }
+    PyDict_SetItemString(
+        EdgeObject_Type.tp_dict, "__dataclass_fields__", default_fields
+    );
 
     base_hash = _EdgeGeneric_HashString("edgedb.Object");
     if (base_hash == -1) {
--- a/edgedb/datatypes/object.c
+++ b/edgedb/datatypes/object.c
@@ -375,7 +375,8 @@ EdgeObject_InitType(void)
         return NULL;
     }
 
-    // Pass the `dataclasses.is_dataclass(obj)` check, the dict is always empty
+    // Pass the `dataclasses.is_dataclass(obj)` check - which then checks
+    // `hasattr(type(obj), "__dataclass_fields__")`, the dict is always empty
     PyObject *default_fields = PyDict_New();
     if (default_fields == NULL) {
         return NULL;
--- a/edgedb/datatypes/object.c
+++ b/edgedb/datatypes/object.c
@@ -189,10 +189,6 @@ object_hash(EdgeObject *o)
 static PyObject *
 object_getattr(EdgeObject *o, PyObject *name)
 {
-    // Used in `dataclasses.as_dict()`
-    if (PyUnicode_CompareWithASCIIString(name, "__dataclass_fields__") == 0) {
-        return EdgeRecordDesc_GetDataclassFields((PyObject *)o->desc);
-    }
     Py_ssize_t pos;
     edge_attr_lookup_t ret = EdgeRecordDesc_Lookup(
         (PyObject *)o->desc, name, &pos);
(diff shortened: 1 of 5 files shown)


i0bs:
Have not tested, but for this part:

```py
return client.query("select User { name, role: { name } }")[0]
```

Will the first index return us the `Role` Python dataclass, or the EdgeQL schema of it?

fantix:
> Have not tested, but for this part:
> 
> ```python
> return client.query("select User { name, role: { name } }")[0]
> ```
> 
> Will the first index return us the `Role` Python dataclass, or the EdgeQL schema of it?

Oh that's the first object in the result set as we are using `query()` instead of `query_single()`.

fantix:
Furthermore, we will be able to separate this into 2 separate files:

`user_role.edgeql`:

```
select User { name, role: { name } }
```

`user_role_edgeql.py`:

```python
import edgedb
import typing
import uuid
from dataclasses import dataclass

@dataclass
class Role:
    id: uuid.UUID
    name: str

    @classmethod
    def __get_validators__(cls):
        return []

@dataclass
class User:
    id: uuid.UUID
    name: str
    role: Role

    @classmethod
    def __get_validators__(cls):
        return []

def query(client: edgedb.Client) -> typing.Sequence[User]:
    return client.query("select User { name, role: { name } }")
```

So that we can generate `user_role_edgeql.py` from `user_role.edgeql` with a console command.

i0bs:
> Oh that's the first object in the result set as we are using `query()` instead of `query_single()`.
> 
> p.s.: Ideally we could do something like:
> 
> ```python
> @app.get("/", response_model=typing.Iterable[User])
> def read_root():
>     return client.query("select User { name, role: { name } }")
> ```
> 
> But we're blocked by [tiangolo/fastapi#3913](https://github.com/tiangolo/fastapi/pull/3913) - so using a regular `list` is the current alternative.

That would be preferred, or you could subsequently use `typing.Mapping` to make a relationship between the dataclass and then client-defined typings. Would that also work?

fantix:
> subsequently use `typing.Mapping` to make a relationship between the dataclass and then client-defined typings. Would that also work?

An example please? Sorry I'm not sure I'm 100% following.

i0bs:
> An example please? Sorry I'm not sure I'm 100% following.

`typing.Mapping` works [like a container,](https://docs.python.org/3/glossary.html#term-mapping) you essentially would have an object that returns a value type based on the key type:

```py
class CustomMapping(typing.Generic[KV, KT]):
    def __getitem__(self, key: KT) -> VT:
        ...
```

This would only be useful if you have custom-defined typings in the client library. For example, [you could use type variables](https://docs.python.org/3/library/typing.html#typing.TypeVar) to represent types from EdgeQL.

edgedb/datatypes/record_desc.c
```diff
@@ -30,6 +30,7 @@ record_desc_dealloc(EdgeRecordDescObject *o)
     PyObject_GC_UnTrack(o);
     Py_CLEAR(o->index);
     Py_CLEAR(o->names);
+    Py_CLEAR(o->get_dataclass_fields);
```
1st1, line 33: Do we need to visit it in `record_desc_traverse`?

edgedb/datatypes/record_desc.c
```diff
@@ -177,12 +178,23 @@ record_desc_dir(EdgeRecordDescObject *o, PyObject *args)
 }
 
 
+static PyObject *
+record_set_dataclass_fields_func(EdgeRecordDescObject *o, PyObject *arg)
+{
+    o->get_dataclass_fields = arg;
```
1st1, line 184: Py_CLEAR it first before the assignment?

1st1, line 184:
```suggestion
    o->get_dataclass_fields_func = arg;
```

edgedb/datatypes/record_desc.c
1st1: OK, this is a function.. the name is confusing. I suggest renaming to `get_dataclass_fields_func`

edgedb/datatypes/object.c
```diff
@@ -189,6 +189,10 @@ object_hash(EdgeObject *o)
 static PyObject *
 object_getattr(EdgeObject *o, PyObject *name)
 {
+    // Used in `dataclasses.as_dict()`
+    if (PyUnicode_CompareWithASCIIString(name, "__dataclass_fields__") == 0) {
+        return EdgeRecordDesc_GetDataclassFields((PyObject *)o->desc);
+    }
```
1st1, line 195: I think this should be moved into the `L_NOT_FOUND` case. Pointless to call this on every other attribute lookup.

edgedb/datatypes/object.c
```diff
@@ -365,6 +369,15 @@ EdgeObject_InitType(void)
         return NULL;
     }
 
+    // Pass the `dataclasses.is_dataclass(obj)` check, the dict is always empty
+    PyObject *default_fields = PyDict_New();
+    if (default_fields == NULL) {
+        return NULL;
+    }
+    PyDict_SetItemString(
+        EdgeObject_Type.tp_dict, "__dataclass_fields__", default_fields
```
1st1, line 378: Why is this needed if we have a getter?

fantix: In order to pass the `dataclasses.is_dataclass(obj)` check, which checks `hasattr(type(obj), "__dataclass_fields__")`

1st1: > which checks hasattr(type(obj), "__dataclass_fields__")

Can you add a comment about that?

1st1 (approved):

edgedb/datatypes/object.c
fantix: Sure - the comment is more complete now

PR #360, opened 2022-08-15
[PoC] Codegen

fantix:
Put `*.edgeql` files under project directory, and run:

```
python -m edgedb.codegen
```

Generated code would require #359 to run properly in FastAPI.

This PR generates JSON Schema from `Parse` response as an intermediate result, using codec instructs copied from the EdgeDB server, so that we could easily migrate to a future codegen frontend. The JSON Schema looks like this (example from #359):

```json
{"$schema": "https://json-schema.org/draft/2020-12/schema",
 "definitions": {"role": {"properties": {"id": {"type": "std::uuid"},
                                         "name": {"type": "string"}},
                          "type": "object"},
                 "user": {"properties": {"id": {"type": "std::uuid"},
                                         "name": {"type": "string"},
                                         "role": {"$ref": "#/definitions/role"}},
                          "type": "object"}},
 "properties": {"input": {"type": "null"},
                "output": {"items": {"$ref": "#/definitions/user"},
                           "type": "array"}},
 "type": "object"}
```

### DIFF
--- a/edgedb/codegen/async_query.py.template
+++ b/edgedb/codegen/async_query.py.template
@@ -1,10 +1,15 @@
-import edgedb
+from __future__ import annotations
+{gen.imports}import edgedb
 
 
-async def {stem}(client: edgedb.AsyncIOClient, *args, **kwargs):
+{gen.aliases}{gen.definitions}async def {stem}(
+    client: edgedb.AsyncIOClient,
+    *args,
+    **kwargs,
+) -> {out_type}:
     return await client.query(
         """\
-        {content}\
+        {query}\
         """,
         *args,
         **kwargs,
--- a/edgedb/codegen/async_query.py.template
+++ b/edgedb/codegen/async_query.py.template
@@ -1,7 +1,7 @@
 import edgedb
 
 
-async def query(client: edgedb.AsyncIOClient, *args, **kwargs):
+async def {stem}(client: edgedb.AsyncIOClient, *args, **kwargs):
     return await client.query(
         """\
         {content}\
--- a/edgedb/codegen/async_query.py.template
+++ b/edgedb/codegen/async_query.py.template
@@ -1,0 +1,11 @@
+import edgedb
+
+
+async def query(client: edgedb.AsyncIOClient, *args, **kwargs):
+    return await client.query(
+        """\
+        {content}\
+        """,
+        *args,
+        **kwargs,
+    )
--- a/edgedb/codegen/async_query.py.template
+++ b/edgedb/codegen/async_query.py.template
@@ -1,16 +1,12 @@
 from __future__ import annotations
-{gen.imports}import edgedb
+{gen.imports}
 
 
 {gen.aliases}{gen.definitions}async def {stem}(
     client: edgedb.AsyncIOClient,
-    *args,
-    **kwargs,
-) -> {out_type}:
+{in_type_args}){out_type}:
     return await client.query(
         """\
         {query}\
         """,
-        *args,
-        **kwargs,
-    )
+{in_type_call}    )
--- a/edgedb/codegen/sertypes.py
+++ b/edgedb/codegen/sertypes.py
@@ -204,25 +204,36 @@ def parse(typedesc: bytes) -> TypeDesc:
 
 
 def describe(
-    desc: TypeDesc, name: str, cardinality: bytes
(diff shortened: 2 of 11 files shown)


edgedb/codegen/async_query.py.template
```diff
@@ -0,0 +1,12 @@
...
+
+
+{gen.aliases}{gen.definitions}async def {stem}(
+    client: edgedb.AsyncIOClient,
+{in_type_args}){out_type}:
+    return await client.query(
+        """\
+        {query}\
+        """,
+{in_type_call}    )
```
1st1, line 12: The formatting here looks a bit weird, is that on purpose? Can you show an example of generated code?

edgedb/codegen/sertypes.py
```diff
@@ -0,0 +1,482 @@
...
+
+    elif t[0] >= 0x80 and t[0] <= 0xFF:
+        # Ignore all type annotations.
+        desc.read_len32_prefixed_bytes()
+        return None
+
+    else:
+        raise NotImplementedError(
+            f"no codec implementation for EdgeDB data class {t}"
+        )
```
1st1, line 192: This looks sufficiently similar to how the codecs tree is built and that code is already in this package. Can we potentially reuse it instead of copy/pasting this?

1st1: You can traverse the tree of codecs and use singledispatch to implement custom logic needed here. IDK, I'd do that if that's not too hard; otherwise this is fine too (albeit this copy/paste still looks worrisome).

edgedb/protocol/protocol.pyx
```diff
@@ -326,6 +326,73 @@ cdef class SansIOProtocol:
 
         return cardinality, in_dc, out_dc, capabilities
 
+    async def raw_parse(self, query: str):
```
1st1, line 329: Why not using the `_parse` method?

edgedb/codegen/async_query.py.template
fantix:
```python
from __future__ import annotations
import dataclasses
import edgedb
import typing
import uuid


TryEnum0 = str


@dataclasses.dataclass
class TryEnum1:
    id: uuid.UUID

    @classmethod
    def __get_validators__(cls):
        return []


def try_enum(
    client: edgedb.Client,
    a: TryEnum0,
    b: edgedb.Object,
) -> typing.Tuple[TryEnum0, TryEnum1]:
    return client.query(
        """\
        select (<MyString>$a, <User>$b)\
        """,
        a=a,
        b=b,
    )
```

Schema:

```
    scalar type MyString extending str;

    type User {...}
```

edgedb/codegen/sertypes.py
fantix: This copy/paste was intended to be deleted after we can do the same in the server-side `sertypes.py` as a language server in the (far) future. This is also why the `raw_parse()` is added below so that we can "recover" the `sertypes.py` codecs.

edgedb/codegen/async_query.py.template
1st1: OK, looks nice!

edgedb/codegen/sertypes.py
1st1: > This is also why the raw_parse() is added below so that we can "recover" the sertypes.py codecs.

I don't understand what do you mean by "recover" here...

fantix: Oh sorry, I meant in the future language server, we won't need to serialize and deserialize the type descriptors, just 1) compile/analyze the query and get `sertypes.py` codecs objects in memory, 2) generate the JSON schema directly using the same code as we are having in this PR here in `sertypes.py`. However we don't have the language server now, so I'm simply copying over part of the `sertypes.py` that re-loads ("recover") the codec objects from serialized type descriptors, and do the (2) step as what we will do in the future language server. An up-side is that we don't have to interfere/hack too much into the C code in edgedb-python now (so that it'll be easier to migrate to the language server in the future), while the downside is to maintain two copies of `sertypes.py` until the language server is developed (but I guess we don't usually modify the codes, so...).

edgedb/protocol/protocol.pyx
fantix: Answered in the thread above :)

fantix:
Close in favor of #363

PR #363, opened 2022-09-05
[WIP] Codegen with the describe() API

fantix:


edgedb/base_client.py
```diff
@@ -287,6 +287,25 @@ async def _execute(self, execute_context: abstract.ExecuteContext) -> None:
                 ),
             )
 
+    async def describe(
+        self, describe_context: abstract.DescribeContext
+    ) -> abstract.DescribeResult:
+        cardinality, in_dc, out_dc, _ = await self._protocol._parse(
```
fantix, line 293: `capabilities` is ignored for now

edgedb/base_client.py
```diff
@@ -287,6 +287,25 @@ async def _execute(self, execute_context: abstract.ExecuteContext) -> None:
...
+            describe_context.query,
+            reg=protocol.CodecsRegistry(),
```
fantix, line 295: `describe()` has no codecs cache for now

edgedb/base_client.py
```diff
@@ -287,6 +287,25 @@ async def _execute(self, execute_context: abstract.ExecuteContext) -> None:
...
+            inline_typenames=True,
+            inline_typeids=True,
```
fantix, line 297: These are actually not correct - they only add type names and IDs into the `Execute` result not the `Parse` result (an exception is non-builtin scalars), so we'll probably need a new `CompilationFlag` (even though it doesn't affect compilation, but there's no other fields in the `Parse` message to set this, unless we change the protocol) in order to include the type names in the `Parse` result.

edgedb/abstract.py
```diff
@@ -48,6 +55,48 @@ class ExecuteContext(typing.NamedTuple):
...
+        raise NotImplementedError
+
+
+class SubType:
+    codec: AbstractCodec
+    name: typing.Optional[str]
+    cardinality: typing.Optional[enums.Cardinality]
+    is_implicit: bool
+    is_link_property: bool
+    is_link: bool
```
fantix, line 91: This is actually defined in `edgedb/introspect.py` already, reusing somehow is preferred.

edgedb/base_client.py
fantix: We'll use introspection query to get all names.

edgedb/datatypes/array.c
```diff
@@ -222,6 +222,36 @@ array_repr(EdgeArrayObject *o)
...
+            return NULL;
+        }
+        PyList_SET_ITEM(y, i, item);
+    }
+    return y;
+}
+
+
+static PyMethodDef array_methods[] = {
+    {"__deepcopy__", (PyCFunction)array_deepcopy, METH_O, NULL},
```
fantix, line 250: Alternatively, we could also implement `__reduce__` here.

edgedb/base_client.py
1st1: That's OK

edgedb/codegen/generator.py
```diff
@@ -0,0 +1,422 @@
...
+            for sub_type in codec.sub_types:
+                if sub_type.is_implicit and sub_type.name != "id":
+                    continue
+                name = f"{rv}{sub_type.name.title()}"
+                # if sub_type.cardinality == edgedb.Cardinality.MANY:
+                #     name = f"{name}Item"
+                #     st_code = self._generate_code(sub_type.codec, name)
+                #     self._imports.add("typing")
+                #     st_code = f"typing.Sequence[{st_code}]"
+                # else:
```
1st1, line 369: let's not commit commented out code

edgedb/datatypes/array.c
```diff
@@ -222,6 +222,36 @@ array_repr(EdgeArrayObject *o)
 }
 
 
+static PyObject *
+array_deepcopy(EdgeArrayObject *x, PyObject *arg)
```
1st1, line 226: Curious why do we need deepcopy support?

edgedb/datatypes/object.c
```diff
@@ -338,8 +338,44 @@ object_dir(EdgeObject *o, PyObject *args)
...
+        if (PyDict_SetItem(y, name, item)) {
+            Py_DECREF(name);
+            Py_DECREF(item);
+            Py_DECREF(y);
+            return NULL;
+        }
+        Py_DECREF(name);
+        Py_DECREF(item);
+    }
+    return y;
```
1st1, line 372: I think it's somewhat a violation that we return vanilla Python types when our types get deepcopied. Why?

edgedb/base_client.py
fantix: Now included in `DescribeResult`.

1st1:
Can you add a comment to the top of generated file like `# autogenerated with {command}`

edgedb/abstract.py
```diff
@@ -47,6 +71,19 @@ class ExecuteContext(typing.NamedTuple):
     state: typing.Optional[options.State]
 
 
+class DescribeContext(typing.NamedTuple):
```
tailhook, line 74: Now these has to be dataclasses, right? Since we don't need to support earlier python versions.

fantix: Why? They are for internal use only and I think named tuples are slightly more efficient? Sorry I didn’t get your point.

tailhook: Ah, well, yes. DescribeContext is internal. But DescribeResult isn't (when `describe_query` becomes public).

Generally point is: `NamedTuple` are the old "poor man's dataclass". If there is no backwards compatibility concerns and no need to compare with tuple, it's better to use dataclass. For example, adding a field into dataclass in the future version is more expected (while for named tuple, different tuple length is technically a breaking change).

And I don't think performance matters here (I also believe that eventually dataclasses should be in par with namedtuples with regards to performance).

tests/codegen/test-project2/generated_async_edgeql.py.assert
```diff
@@ -0,0 +1,362 @@
...
+        select <int64>$arg_name_with_underscores\
+        """,
+        arg_name_with_underscores=arg_name_with_underscores,
+    )
+
+
+async def select_args(
+    client: edgedb.AsyncIOClient,
+    *,
+    str: str,
```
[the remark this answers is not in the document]
fantix: Ah, right, thanks! Let me fix it

tests/codegen/test-project2/object/link_prop_async_edgeql.py.assert
```diff
@@ -0,0 +1,62 @@
...
+    name: str
+    friends: typing.List[LinkPropResultFriendsItem]
+
+
+@dataclasses.dataclass
+class LinkPropResultFriendsItem(NoPydanticValidation):
+    id: uuid.UUID
+    name: str
+
+    def __getitem__(self, key: str) -> typing.Any:
```
1st1, line 34: I thought we'd generate proper typed overloads here?

fantix: Right, this specific `assert` is for Python 3.7 which doesn't have `typing.Literal`

1st1: Ah, ok, i figured it out -- there's another properly typed .assert

1st1 (approved):
Epic work Fantix. I think this is ready.

PR #365, opened 2022-09-12
Ping first if conn is idle for too long

fantix:
The blocking sockets don't have active connection_lost() events, if the user is trying to execute a non-read-only query over a disconnected connection, is_closed() cannot capture the disconnection and the retry loop wouldn't work. In this case, we are sending a SYNC before the query, so that we could capture the connection error and reconnect.

This is tailored to handle the server-side `session_idle_timeout`; for regular network interruptions within the `session_idle_timeout`, we don't want to sacrifice performance by inserting more frequent SYNCs just to avoid connection errors on non-read-only queries.

### DIFF
--- a/edgedb/blocking_client.py
+++ b/edgedb/blocking_client.py
@@ -17,6 +17,7 @@
 #
 
 
+import datetime
 import queue
 import socket
 import ssl
@@ -32,8 +33,13 @@ from . import transaction
 from .protocol import blocking_proto
 
 
+DEFAULT_SESSION_IDLE_TIMEOUT = datetime.timedelta(seconds=60)
+DEFAULT_PING_BEFORE_IDLE_TIMEOUT = datetime.timedelta(seconds=5)
+MINIMUM_PING_WAIT_TIME = datetime.timedelta(seconds=1)
+
+
 class BlockingIOConnection(base_client.BaseConnection):
-    __slots__ = ()
+    __slots__ = ("_ping_wait_time",)
 
     async def connect_addr(self, addr, timeout):
         deadline = time.monotonic() + timeout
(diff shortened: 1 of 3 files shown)


edgedb/blocking_client.py
```diff
@@ -97,6 +103,17 @@ async def connect_addr(self, addr, timeout):
 
             self._protocol = proto
             self._addr = addr
+            self._ping_wait_time = max(
+                (
+                    getattr(
+                        self.get_settings().get("system_config"),
+                        "session_idle_timeout",
+                        DEFAULT_SESSION_IDLE_TIMEOUT,
```
1st1, line 111: I don't remember -- can it be missing or we always send it? (I think we always do, no?)

fantix: Yeah, I also think we always send:

https://github.com/edgedb/edgedb/blob/60c07267b310bffbf9a827635b1d51be4a71d72a/edb/server/bootstrap.py#L1131-L1141

then

https://github.com/edgedb/edgedb/blob/60c07267b310bffbf9a827635b1d51be4a71d72a/edb/server/protocol/binary.pyx#L536-L539

1st1: Then there's no need for DEFAULT_SESSION_IDLE_TIMEOUT?

fantix: Right!

fantix:
Actually Elvis suggested over the meeting that, a proper fix should be that the server also sends a `Terminate` message before shutting down (idle) connections, so that the client won't have a vague state about the disconnection. I'll also do that, but this PR I think is still needed for EdgeDB 2.2 and lower.

tailhook:
1. On unixes you can do just `poll` with `POLLOUT`. If OS thinks connection is down, it will return an error. I think you can achieve the same with `select` on windows.
2. As `Terminate` is a subject to race condition, you have to handle it in every receive loop, which is quite inconvenient. And TCP RST packet sent by OS (which is caught by (1)) is almost as reliable as application specific `Terminate` packet.
3. `Sync` is probably the only reliable way (it covers more than (1) and (2) combined), although costly. But that cost is probably not critical as it is only incurred when connection was idle for the long time.

Also we could probably skip `Sync` if we send `Parse` instead of cached query, but that is probably over-optimizing too niche case.

I would go with (1) or (3), the latter is in current PR. As I don't see a significant difference between (1) and (2), whereas (2) complicates client implementation.

1st1 (approved):

1st1:
> Sync is probably the only reliable way

Yeah.

PR #366, opened 2022-09-14
Redo edgedb basic types to inherit from builtin types

fantix:
- [x] `edgedb.Tuple` is `tuple`
   * `weakref.ref()` will stop working on `edgedb.Tuple` objects.
- [x] `edgedb.NamedTuple` is a subclass of `tuple`
   * Actual named tuple objects will be an instance of a transient heap type `DerivedNamedTuple` (subclass of `edgedb.NamedTuple`)
   * `weakref.ref()` will stop working on `edgedb.NamedTuple` objects.
- [x] `edgedb.Set` is `list`
- [x] `edgedb.Array` is `list`
- [x] `edgedb.EnumValue` is a subclass of `enum.Enum`
- [x] link and linkprop? We'll keep them as they are now, and they are not included in the `dataclasses.asdict()` result.
- [x] docs
- [x] tests

Benchmark:

2.3 GHz Intel 4 Core, Turbo Boost Off, concurrency=10, time=100s

image
image

datatypes branch (left):
```
queries:	184141
qps:		1841 q/s
min latency:	1.88ms
avg latency:	5.42ms
max latency:	28.36ms
```

0.24 (right):
```
queries:	188075
qps:		1880 q/s
min latency:	1.37ms
avg latency:	5.30ms
max latency:	88.56ms
```


i0bs:
https://github.com/edgedb/edgedb-python/blob/dfb8c8b009838eec44a312d467edd4782197897f/edgedb/datatypes/array.c#L47

Will other types from stdlib such as arrays inherit from builtin?

fantix:
> Will other types from stdlib such as arrays inherit from builtin?

Yes, but array may just be a subclass of `tuple` because it's immutable.

i0bs:
> Yes, but array may just be a subclass of `tuple` because it's immutable.

Has it always been immutable? I thought EdgeQL supported [mutation to the shape and size of arrays.](https://www.edgedb.com/docs/edgeql/update)

fantix:
> Has it always been immutable?

Oh this PR is the edgedb-python construct of arrays only in query results, it was and will be immutable.

1st1:
Why can't we use the builtin `tuple` type for our tuples? The `__repr__` is basically compatible, so maybe just drop our custom tuple implementation and use the builtin one? It has a freelist optimization.

edgedb/datatypes/tuple.c
```diff
@@ -51,16 +45,14 @@ EdgeTuple_New(Py_ssize_t size)
         return NULL;
     }
 
-    EdgeTupleObject *obj = NULL;
+    PyTupleObject *obj = NULL;
 
-    EDGE_NEW_WITH_FREELIST(EDGE_TUPLE, EdgeTupleObject,
+    EDGE_NEW_WITH_FREELIST(EDGE_TUPLE, PyTupleObject,
```
1st1, line 50: Basically we don't need this line. Please double check, but IIRC the builtin tuple type already uses freelists.

edgedb/datatypes/tuple.c
```diff
@@ -70,37 +62,34 @@ int
...
-tuple_hash(EdgeTupleObject *o)
+tuple_hash(PyTupleObject *o)
 {
     return _EdgeGeneric_Hash(o->ob_item, Py_SIZE(o));
 }
 
 
 static int
-tuple_traverse(EdgeTupleObject *o, visitproc visit, void *arg)
+tuple_traverse(PyTupleObject *o, visitproc visit, void *arg)
```
1st1, line 92: And if we use the builtin type we'd no longer have to reimplement these.

edgedb/datatypes/tuple.c
fantix: Yes it does: https://github.com/python/cpython/blob/3.7/Objects/tupleobject.c#L87-L110

edgedb/datatypes/namedtuple.c
```diff
@@ -338,35 +351,103 @@ static PyMethodDef namedtuple_methods[] = {
...
+    assert(Py_TYPE(type)->tp_itemsize > sizeof(PyObject *));
+    rv = PyObject_GC_Resize(PyObject, type, size  + 1);
+    if (rv == NULL) {
+        PyObject_GC_Del(type);
+        return NULL;
+    }
+    Py_SIZE(rv) = size;
+    EdgeNamedTuple_Type_DESC(rv) = desc;
+    // desc is also stored in tp_dict for refcount.
+    PyDict_SetItemString(((PyTypeObject *)rv)->tp_dict, "__desc__", desc);
```
fantix, line 419: Because the Stable ABI doesn't have the `PyHeapTypeObject` type, we're using this hack to store a fast access for the `desc` pointer.

edgedb/datatypes/namedtuple.c
```diff
@@ -58,62 +60,34 @@ EdgeNamedTuple_New(PyObject *desc)
         return NULL;
     }
 
-    EdgeNamedTupleObject *nt = NULL;
-    EDGE_NEW_WITH_FREELIST(EDGE_NAMED_TUPLE, EdgeNamedTupleObject,
-                           &EdgeNamedTuple_Type, nt, size);
+    PyTupleObject *nt = NULL;
+    EDGE_NEW_WITH_FREELIST(EDGE_NAMED_TUPLE, PyTupleObject, type, nt, size);
```
fantix, line 64: We cannot reuse the builtin freelist for `tuple` because the `ob_type` is not `&PyTuple_Type`.

edgedb/datatypes/namedtuple.c
```diff
@@ -58,62 +60,34 @@ EdgeNamedTuple_New(PyObject *desc)
...
     assert(nt != NULL);
-    assert(EdgeNamedTuple_Check(nt));
+    if (Py_TYPE(nt) != type) {
+        Py_DECREF(Py_TYPE(nt));
+        Py_INCREF(type);
+        Py_TYPE(nt) = type;
+    }
```
fantix, line 70: When we are switching in and out of the freelist, make sure the `ob_type` is still what we expected.

edgedb/datatypes/namedtuple.c
fantix: It turns out that we cannot blindly resize a heap type object, because `PyType_Ready()` sets several pointers like `__mro__` and subclass weakrefs in base classes, the address of the type object is not expected to change after that.

edgedb/datatypes/namedtuple.c
fantix: This is now done in a different way - we always DECREF the type before putting memory into the freelist, and set and INCREF the type when resuing memory.

fantix:
> > Has it always been immutable?
> 
> Oh this PR is the edgedb-python construct of arrays only in query results, it was and will be immutable.

UPDATE: we decided to use plain Python `list` for `edgedb.Set` and `edgedb.Array`, so the query results will be partially mutable, but again that's just in user memory.

i0bs:
> UPDATE: we decided to use plain Python list for edgedb.Set and edgedb.Array, so the query results will be partially mutable, but again that's just in user memory.

Does the builtin list object support the same optimisation as before? Does this improve querying perf?

fantix:
> > UPDATE: we decided to use plain Python list for edgedb.Set and edgedb.Array, so the query results will be partially mutable, but again that's just in user memory.
> 
> Does the builtin list object support the same optimisation as before? Does this improve querying perf?

The main optimization is still here, which is using the C API `PyList_SET_ITEM` for fast initialization, but we lost the (cached) hashing function, which is not a critical use. Also we lost the freelist of array, which may have slight performance impact when arrays are frequently (re-)allocated. Other than that, query performance should not be affected, I’ll run benchmarks to test it.

i0bs:
I'd love to hear more about the benchmarks once they're available. This sounds like a great change if memory allocation doesn't prove to be too impactful for builtin inheritance. Great stuff @fantix ❤️

edgedb/datatypes/namedtuple.c
```diff
@@ -58,64 +61,72 @@ EdgeNamedTuple_New(PyObject *desc)
...
+        ) {
+            PyErr_NoMemory();
+            return NULL;
+        }
+        nt = PyObject_GC_NewVar(PyTupleObject, type, size);
+        if (nt == NULL) {
+            return NULL;
+        }
+#if PY_VERSION_HEX < 0x03080000
+        // Workaround for Python issue 35810; no longer necessary in Python 3.8
```
1st1, line 94: ouch!

edgedb/datatypes/namedtuple.c
```diff
@@ -58,64 +61,72 @@ EdgeNamedTuple_New(PyObject *desc)
...
-namedtuple_traverse(EdgeNamedTupleObject *o, visitproc visit, void *arg)
+namedtuple_traverse(PyTupleObject *o, visitproc visit, void *arg)
 {
-    Py_VISIT(o->desc);
-
-    Py_ssize_t i;
-    for (i = Py_SIZE(o); --i >= 0;) {
+#if PY_VERSION_HEX >= 0x03090000
+    // This was not needed before Python 3.9 (Python issue 35810 and 40217)
+    Py_VISIT(Py_TYPE(o));
```
1st1, line 127: ouch!!

edgedb/datatypes/namedtuple.c
```diff
@@ -220,11 +237,88 @@ namedtuple_new(PyTypeObject *type, PyObject *args, PyObject *kwargs) {
 
 
 static PyObject *
-namedtuple_getattr(EdgeNamedTupleObject *o, PyObject *name)
+namedtuple_derived_new(PyTypeObject *type, PyObject *args, PyObject *kwargs) {
```
1st1, line 240: What is this function used for?

Will it be used with `a_fetched_named_tuple.__class__((1, 2, 3))`?

edgedb/datatypes/hash.c
```diff
@@ -41,9 +41,8 @@
 Py_hash_t
 _EdgeGeneric_Hash(PyObject **els, Py_ssize_t len)
 {
-    /* Python's tuple hash algorithm.  Hashes of edgedb.Tuple and
-       edgedb.NamedTuple must be equal to hashes of Python't tuples
-       with the same elements */
+    /* Python's tuple hash algorithm.  Hashes of edgedb.NamedTuple must be
+       equal to hashes of Python't tuples with the same elements */
```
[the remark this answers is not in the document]
fantix: Oh this is a wrong comment - this is still needed (and only) by the object data type. I'll change the comments.

edgedb/datatypes/namedtuple.c
fantix: Yes, in `dataclasses.py`:

```python
    elif isinstance(obj, tuple) and hasattr(obj, '_fields'):
        return type(obj)(*[_asdict_inner(v, dict_factory) for v in obj])
```

edgedb/datatypes/datatypes.pxd
```diff
@@ -47,46 +46,35 @@ cdef extern from "datatypes.h":
...
     object EdgeNamedTuple_New(object)
-    int EdgeNamedTuple_SetItem(object, Py_ssize_t, object) except -1
+    object EdgeNamedTuple_Type_New(object)
 
     object EdgeObject_InitType()
     object EdgeObject_New(object);
     int EdgeObject_SetItem(object, Py_ssize_t, object) except -1
     object EdgeObject_GetRecordDesc(object)
 
-    bint EdgeSet_Check(object)
```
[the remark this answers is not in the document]
fantix: Sure thing, that makes sense!

docs/api/types.rst
```diff
@@ -287,3 +255,30 @@ DateDuration
...
+=========
+
+.. py:class:: EnumValue()
+
+    An immutable value representing an EdgeDB enum value.
+
+    ``edgedb.EnumValue`` is a subclass of :py:class:`enum.Enum <python:enum.Enum>`
+    since version 0.25.0. Actual enum values are instances of ad-hoc classes
+    created by the codec to represent the specific enums, but the ad-hoc
+    classes are direct subclasses of ``edgedb.EnumValue``.
```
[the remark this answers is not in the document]
fantix: Added a test 🤞

edgedb/datatypes/hash.c
```diff
@@ -24,79 +24,6 @@
 #include "datatypes.h"
 
 
-#if PY_VERSION_HEX >= 0x03080000
```
1st1, line 27: You can also now remove the "Portions Copyright 2001-2018" notice from the top comment in this file (it was there for the hashing function)

1st1 (approved):
LGTM! (Aside from one little nit, see the new comment)

fantix:
@i0bs benchmark updated

PR #379, opened 2022-10-17
Deprecate object links and simplify link property access

fantix:
Accessing `edgedb.Link` or `edgedb.LinkSet` on `edgedb.Object` instances through `__getitem__()` will now raise a `DeprecationWarning`. Accessing link properties through links will be dropped in a future version. Instead, link properties can be accessed now through `edgedb.Object.__getitem__(linked_obj, '@link_prop_name')`, or in short, `linked_obj['@link_prop_name']`.

### DIFF
--- a/edgedb/datatypes/object.c
+++ b/edgedb/datatypes/object.c
@@ -216,6 +216,25 @@ static PyObject *
 object_getitem(EdgeObject *o, PyObject *name)
 {
     Py_ssize_t pos;
+    int prefixed = 0;
+    if (PyUnicode_Check(name)) {
+        PyObject *at_sign = PyUnicode_FromString("@");
+        if (at_sign == NULL) {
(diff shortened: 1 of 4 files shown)


edgedb/datatypes/object.c
```diff
@@ -216,28 +216,83 @@ static PyObject *
 object_getitem(EdgeObject *o, PyObject *name)
 {
     Py_ssize_t pos;
+    int prefixed = 0;
+    if (PyUnicode_Check(name)) {
+        PyObject *at_sign = PyUnicode_FromString("@");
```
1st1, line 221: Can you cache this object globally? (Only if it's easy and we can use some safe CPython API for global module state)

1st1 (approved):

1st1 (approved):

PR #380, opened 2022-10-18
Add codegen docs

colinhacks:


docs/api/codegen.rst
```diff
@@ -0,0 +1,90 @@
...
+.. py:currentmodule:: edgedb
+
+The ``edgedb-python`` package exposesa a command-line tool to generate
+typesafe functions using dataclasses from ``*.edgeql`` files.
+
+.. code-block:: bash
+
+  $ edgedb-py
+
+Consider a simple query that lives in a function called ``get_number.py``:
```
fantix, line 16: should be `get_number.edgeql`?

docs/api/codegen.rst
```diff
@@ -0,0 +1,90 @@
...
+~~~~~~
+
+By default, the generated code uses an ``async`` API. The generator supports
+additional targets via the ``--target`` flag.
+
+.. code-block:: bash
+
+  $ edgedb-py --target async        # generate async function (default)
+  $ edgedb-py --target blocking     # generate blocking code
+  $ edgedb-py --target pydantic     # use pydantic typing
```
fantix, line 53: This is more like:

```suggestion
  $ edgedb-py --target pydantic     # skip validation for pydantic if installed
```

I should probably rename this target a bit 🤔

docs/api/codegen.rst
```diff
@@ -0,0 +1,90 @@
...
+
+The names of the generated files will differ accordingly:
+``{query_filename}_{target}_edgeql.py``.
```
fantix, line 56: Oh in this case, the `pydantic` is a special target that doesn't really change the generated file name. I think I should probably use a CLI option like `--skip-pydantic-validation` rather than an option.

fantix (commented):
Thanks for adding this! Looks great, it's just I should probably update the codegen PR first.

fantix (approved):

PR #386, opened 2022-10-22
add atomic iteration option to codegen

TomFaulkner:
Adds atomic transactions option to change client to transaction iteration.

This and similar for blocking:
```python
if self._atomic:
    print(f"{INDENT}transaction: edgedb.asyncio_client.AsyncIOIteration,", file=buf)
else:
    print(f"{INDENT}client: edgedb.AsyncIOClient,", file=buf)
```

### DIFF
--- a/edgedb/codegen/cli.py
+++ b/edgedb/codegen/cli.py
@@ -52,6 +52,11 @@ parser.add_argument(
     default=["async"],
     help="Choose one or more targets to generate code (default is async)."
 )
+parser.add_argument(
+    "--atomic",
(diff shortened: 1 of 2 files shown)


fantix:
Thanks for the PR! I'm thinking we should probably do something like this instead:

```python
async def my_query(
    executor: edgedb.AsyncIOExecutor,
    ...
) -> MyQueryResult:
    return await executor.query_single(...)
```

So that both `AsyncIOClient` and `AsyncIOIteration` (transaction) are supported.

PR #399, opened 2022-11-17
Output pretty error if possible

### LINKED ISSUE #395: Python client not showing the exact location of the syntax error
```
...
    cardinality, in_dc, out_dc, capabilities = await self._protocol._parse(
  File "edgedb/protocol/protocol.pyx", line 321, in _parse
edgedb.errors.QueryError: missing a type cast before the parameter

```

This is really frustrating especially for newbies in EdgeQL like me

fantix:
Error position and hint are now by default included if present (managed by `EDGEDB_ERROR_HINT`), colored if the stderr refers to a terminal, overridden by `EDGEDB_PRETTY_ERROR`.

Fixes #395 

```python
c.query("select $0")
```

image

```python
c.query("select ('\033[95m嘿嘿嘿', 1 < '\033[95m哈哈😁lol');")
```

image

```python
c.query("""\
select (
    '\033[95m哈哈哈哈哈哈哈😁', '\033[95m嘿嘿嘿' < (
        2, 3, 4,
    ), 345
);
""")
```

image

### DIFF
--- a/edgedb/errors/_base.py
+++ b/edgedb/errors/_base.py
@@ -212,12 +212,13 @@ def _severity_name(severity):
 
 
 def _format_error(msg, query, start, offset, line, col, hint):
+    _init_colors()
     rv = io.StringIO()
     rv.write(f"{BOLD}{msg}{ENDC}{LINESEP}")
     lines = query.splitlines(keepends=True)
     num_len = len(str(len(lines)))
-    rv.write(f"{OKBLUE}{'':>{num_len}} ┌─{ENDC} query:{line}:{col}{LINESEP}")
-    rv.write(f"{OKBLUE}{'':>{num_len}} │ {ENDC}{LINESEP}")
+    rv.write(f"{BLUE}{'':>{num_len}} ┌─{ENDC} query:{line}:{col}{LINESEP}")
+    rv.write(f"{BLUE}{'':>{num_len}} │ {ENDC}{LINESEP}")
     for num, line in enumerate(lines):
         length = len(line)
         line = line.rstrip()  # we'll use our own line separator
@@ -231,11 +232,11 @@ def _format_error(msg, query, start, offset, line, col, hint):
             first_half = repr(line[:start])[1:-1]
             line = line[start:]
             length -= start
-            rv.write(f"{OKBLUE}{num + 1:>{num_len}} │   {ENDC}{first_half}")
+            rv.write(f"{BLUE}{num + 1:>{num_len}} │   {ENDC}{first_half}")
             start = _unicode_width(first_half)
         else:
             # Multi-line error continues
-            rv.write(f"{OKBLUE}{num + 1:>{num_len}} │ {FAIL}│ {ENDC}")
+            rv.write(f"{BLUE}{num + 1:>{num_len}} │ {FAIL}│ {ENDC}")
 
         if offset > length:
             # Error is ending beyond current line
@@ -243,7 +244,7 @@ def _format_error(msg, query, start, offset, line, col, hint):
             rv.write(f"{FAIL}{line}{ENDC}{LINESEP}")
             if start >= 0:
                 # Multi-line error starts
-                rv.write(f"{OKBLUE}{'':>{num_len}} │ "
+                rv.write(f"{BLUE}{'':>{num_len}} │ "
                          f"{FAIL}╭─{'─' * start}^{ENDC}{LINESEP}")
             offset -= length
             start = -1  # mark multi-line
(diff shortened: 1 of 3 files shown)


edgedb/errors/_base.py
```diff
@@ -198,3 +324,9 @@ def _severity_name(severity):
 EDGE_SEVERITY_ERROR = 120
 EDGE_SEVERITY_FATAL = 200
 EDGE_SEVERITY_PANIC = 255
+
+
+LINESEP = os.linesep
+ENV_ON_FLAGS = {"1", "yes", "y", "true", "t", "on"}
```
1st1, line 330: Should we maybe copy edgedb server flag policy and have an enum here? `auto | enabled | disabled` should work

1st1 (approved):
Overall looks amazing. I didn't thoroughly review the printing logic though, cursory it looks OK. I'd change the way we define valid valued for ENV vars to enums like in server

edgedb/errors/_base.py
```diff
@@ -180,15 +226,83 @@ def _severity_name(severity):
...
+                         f"{c.FAIL}{'^' * size} {hint}{c.ENDC}")
+            else:
+                # End of multi-line error
+                rv.write(f"{c.BLUE}{'':>{num_len}} │ "
+                         f"{c.FAIL}╰─{'─' * (size - 1)}^ {hint}{c.ENDC}")
+            break
+    return rv.getvalue()
+
+
+def _unicode_width(text):
```
msullivan, line 284: Note that this is an approximation that will overestimate the length of strings using combining characters that *don't* normalize away.
This means we can get stuff like
```
cannot redefine property 'x' of object type 'std::FreeObject' as scalar type 'std::str'
  ┌─ query:2:26
  │ 
2 │   select { x := 1 } { x := 'f̷͈͎͒̕ǫ̴̏͌ö̶̱̘' };
  │                            ^^^^^^^^^^^^^^^^ error
```

https://zalgo.org/ has a great generator for pathological unicode strings.

Probably this is almost always fine.

msullivan: The CLI's error formatter does seem to handle this one right. I'm not sure exactly what algorithm it uses.
I think to some extent anything we do is an approximation because it depends on assumptions about how the terminal will render it.

Unicode does have a notion of "grapheme clusters", and there are python libraries for handling them (https://github.com/alvinlindstam/grapheme). I'm not sure that this is the best approach either, though, since the width of those clusters can vary and I don't know if there is good way to guess that well. (Consider `🇨🇦`, which depending on your terminal will show as either a flag or the letters CA, but either way will probably take two characters, but is only one grapheme cluster.)

Maybe the best approach on our side would be basically just to drop combining characters.

There is a potentially infinitely deep rabbit hole here.

tailhook: Rust uses de facto standard in Rust: https://crates.io/crates/unicode-width
Which uses Unicode Standard Annex 11 for calculating width.

Note: unicode width != number of graphemes. As width of a grapheme can be two chars.

And you're right that in generic case it's impossible to calculate width. Because some characters can have width not just depending on the terminal but also on the specific font (if there is a character for a specific set of emojis). `unicode-width`'s frontpage gives an example.

# Dropping Zero-width Joiners

1. I've tested three terminals I have around (xfce4, alacritty, wezterm), and all of them do match width with unicode width, even though some of them do not render some of the characters (i.e. flags are only supported in wezterm). *Disclaimer: alacritty and wezterm are implemented in Rust so this explains it a bit, although xfce4 is not and has much longer history.*
2. Copying from any of the three terminal's I've tried does discard zero-width joiner by itself
3. Rendering in Firefox on my machine gives invalid width regardless of if there is a zero-width joiner (i.e. it somehow draws emojis 1.5 of character width :shrug: )
4. Rendering in Chrome on my machine gives good results when zero-width joiner is dropped (like normally copying from terminal), but doesn't if I somehow manage to keep that character (i.e. piping output to `xsel -b` manually).

Note: zalgo-generated stuff looks like plain `foo` in all three terminals (although those unicode tricks are present in the output)

The summary of this: dropping zero-width joiners is safe to do, although looks unnecessary in my setup. If that's not true for some popular terminals or on other systems it could be done.

# Fixup

For my examples this function works:
```
def unicode_width(s):
    return sum(0 if unicodedata.category(c) in ('Mn', 'Cf') else
               2 if unicodedata.east_asian_width(c) == "W" else 1
        for c in s)
```
But I'm not sure if this is good enough. The libraries doing this (rustic unicode-width, and pythonic urwid) have their own width tables.

fantix: Added in #404

PR #401, opened 2022-11-18
Disallow None in elements of array argument

### LINKED ISSUE #376: Reject Nones in arrays on client side
edgedb-python will currently happily encode None in array arguments as NULL. 
https://github.com/edgedb/edgedb/pull/4511 prevents this on the server side, but probably we should reject it on the client side also?

fantix:
Fixes #376


msullivan (approved):
Looks good.

Note that we also do this check on the server side, but doing it client side is also good

1st1 (approved):

PR #423, opened 2023-03-14
Update for rules of instance names

### LINKED ISSUE #420: Cloud instance name should allow leading numerics
`1st1/test` or `fantix/1st1test` should be accepted as valid instance names.

* Regexp: `^(\w(?:\w|-(?=\w))*)(?:/(\w(?:\w|-(?=\w))*))?$`
* Total length should not be greater than 62 if `/` (inclusive) is present, so that the translated cloud domain label `inst--org` is no longer than 63 letters.
* Don't narrow the limit on local instance names

fantix:
* Instance names allow leading digits
* Cloud instance name max length: 62
* Dashes are allowed, except for consecutive ones like --
* Cloud instance name must not have underscores

Fixes #420

Depends on https://github.com/edgedb/shared-client-testcases/pull/27

### DIFF
--- a/edgedb/con_utils.py
+++ b/edgedb/con_utils.py
@@ -74,7 +74,12 @@ HUMAN_US_RE = re.compile(
     r'((?:(?:\s|^)-\s*)?\d*\.?\d*)\s*(?i:us(\s|\d|\.|$)|microseconds?(\s|$))',
 )
 INSTANCE_NAME_RE = re.compile(
-    r'^([A-Za-z_]\w*)(?:/([A-Za-z_]\w*))?$',
+    r'^(\w(?:\w|-(?=\w))*)(?:/(\w(?:\w|-(?=\w))*))?$',
+    re.ASCII,
+)
+DSN_RE = re.compile(
+    r'^[a-z]+://',
+    re.IGNORECASE,
 )
 
 
@@ -519,11 +524,10 @@ def _parse_connect_dsn_and_args(
 ):
     resolved_config = ResolvedConnectConfig()
 
-    dsn, instance_name = (
-        (dsn, None)
-        if dsn is not None and re.match('(?i)^[a-z]+://', dsn)
-        else (None, dsn)
-    )
+    if dsn and DSN_RE.match(dsn):
+        instance_name = None
+    else:
+        instance_name, dsn = dsn, None
 
     has_compound_options = _resolve_config_options(
         resolved_config,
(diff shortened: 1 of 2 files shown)


edgedb/con_utils.py
```diff
@@ -861,6 +865,12 @@ def _parse_cloud_instance_name_into_config(
     org_slug: str,
     instance_name: str,
 ):
+    label = f"{instance_name}--{org_slug}"
+    if len(label) > 63:
```
elprans, line 869: Let's make this a constant like the regexp.

edgedb/con_utils.py
```diff
@@ -861,6 +865,12 @@ def _parse_cloud_instance_name_into_config(
...
+        raise ValueError(
```
elprans, line 870: Should probably be `errors.InterfaceError` or a subclass.

edgedb/con_utils.py
```diff
@@ -861,6 +865,12 @@ def _parse_cloud_instance_name_into_config(
...
+            f"invalid instance name: too long for cloud: "
```
elprans, line 871:
```suggestion
            f"invalid instance name: cloud instance name length cannot exceed {max} bytes"
```

elprans (approved):

edgedb/con_utils.py
```diff
@@ -74,7 +74,12 @@
     r'((?:(?:\s|^)-\s*)?\d*\.?\d*)\s*(?i:us(\s|\d|\.|$)|microseconds?(\s|$))',
 )
 INSTANCE_NAME_RE = re.compile(
-    r'^([A-Za-z_]\w*)(?:/([A-Za-z_]\w*))?$',
+    r'^(\w(?:\w|-(?=\w))*)(?:/(\w(?:\w|-(?=\w))*))?$',
```
vpetrovykh, line 77: If you're doing `-(?=\w)` you might as well not bother with a look-ahead and instead use the simpler and more explicit approach of actually saying that any `\w` except the first one can be optionally preceded by a single `-`:
`r'^(\w(?:-?\w)*)(?:/(\w(?:-?\w)*))?$'`

It has the same effect of allowing single `-` to be used inside words, but might be a little easier to read for a maintainer (because it uses less obscure regexp features).

edgedb/con_utils.py
fantix: I'll create a new PR for this - too many of them to fix in a CRF.

PR #426, opened 2023-04-21
Add optional default to generated params

### LINKED ISSUE #422: Default to None for generated functions with defaults
**Describe the bug**

The python code generator does not allow the caller to skip optional parameters, as it requires every parameter
to be passed as None for these be skipped.

For example, see the code in the tests: https://github.com/edgedb/edgedb-python/blob/a5067ef013b00b210da3a47ebeed61a155a29ff3/tests/codegen/test-project2/generated_async_edgeql.py.assert#L198-L216

This can make "update" calls somewhat cumbersome, as it requires spelling out all fields that are not being updated as `None`.

Reference discord thread: https://discord.com/channels/841451783728529451/849374935209607199/1081881229676265592

**Reproduction**

Generate code with `edge-py` using a query that includes a number of optional parameters.

**Expected behavior**

I'd like to see something like `type | None = None` which would bring it in-line with how Typescript parameters work: https://www.edgedb.com/docs/clients/js/parameters#optional-parameters

As a workaround, I created a helper function that copies all parameters from the generated function but defaults to None for the optional arguments.

**Versions (please complete the following information):**

- OS: macOS Ventura 13.2.1 (22D68)
- EdgeDB version: 2.12+a5dfd9f
- EdgeDB CLI version:  2.3.1+5d93f42
- `edgedb-python` version: 1.3.0
- Python version: 3.10.6

**Additional context**

Note: I raised another issue related to generated code and enum values here: https://github.com/edgedb/edgedb-python/issues/419

fantix:
Fixes #422


tailhook (approved):
Other than that single nit, looks god.

JesseFarebro:
Looks like this never made it into the 1.x branch as I've ran into this issue using 1.9.0. Are there any plans to merge this into 1.x?

PR #433, opened 2023-05-25
Add savepoint API

### LINKED ISSUE #407: Transaction savepoints are essential
The Python client does not currently support transaction savepoints, but they are an essential feature. For example it's currently impossible to recover from errors (e.g. `CardinalityViolationError`) within transactions - the entire transaction has to be retried. There are additional difficulties arising in test writing and when dealing with external failures without losing (all) progress in a transaction.

fantix:
Please see tests for use case.

- [ ] Add docs

Fixes #407

### DIFF
--- a/tests/test_async_tx.py
+++ b/tests/test_async_tx.py
@@ -34,6 +34,10 @@ class TestAsyncTx(tb.AsyncQueryTestCase):
         };
     '''
 
+    TEARDOWN_METHOD = '''
+        DELETE test::TransactionTest;
+    '''
+
     TEARDOWN = '''
         DROP TYPE test::TransactionTest;
     '''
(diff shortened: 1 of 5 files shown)


tests/test_async_tx.py
```diff
@@ -104,3 +108,54 @@ async def test_async_transaction_exclusive(self):
...
+
+    async def test_async_transaction_savepoint_1(self):
+        async for tx in self.client.transaction():
+            async with tx:
+                sp1 = await tx.declare_savepoint("sp1")
+                sp2 = await tx.declare_savepoint("sp2")
+                with self.assertRaisesRegex(
+                    edgedb.InterfaceError, "savepoint.*already exists"
+                ):
+                    await tx.declare_savepoint("sp1")
```
elprans, line 120: This is contrary to server savepoint semantics, masking an earlier savepoint is allowed:

```
edgedb> start transaction;
OK: START TRANSACTION
edgedb[tx]> declare savepoint sp1;
OK: DECLARE SAVEPOINT
edgedb[tx]> declare savepoint sp1;
OK: DECLARE SAVEPOINT
```

tailhook: This releases previous savepoint with that name, right?

tailhook (commented):
Three questions:
1. Why do we skip context API: `with tx.declare_savepoint("name")` entirely?
2. We should allow skipping name argument and generate something unique (e.g. uuid). Actually while we have variables in python (I mean `sp_name = conn.declare_savepoint()` ) names are useless, will only provoke unexpected conflicts.
3. Do we need `declare_` part? `x = tx.savepoint()` is nice, no?

Context API fits in pretty nicely, even with rollback (since rollback and release aren't mutually exclusive):
```python
with tx.savepoint() as sp:
  # do something
  if was_already_applied:
      sp.rollback()
```

It can be added later, but is pretty uncontroversial I think.

1st1:
> We should allow skipping name argument and generate something unique (e.g. uuid). Actually while we have variables in python (I mean sp_name = conn.declare_savepoint() ) names are useless, will only provoke unexpected conflicts.

+1

> Do we need declare_ part? x = tx.savepoint() is nice, no? (we already do that with transactions by skipping start part)

Yes, please, it should be `tx.savepoint()` with an optional name argument (although I'd not even have the name argument at all, why would you want it?)

tests/test_async_tx.py
1st1: Yeah, I think Paul is right.

fantix:
Pushed a fix to rename `declare_savepoint()` to `savepoint()` and dropped user-defined names (always use s + uuid instead).

> (Context API) can be added later, but is pretty uncontroversial I think.

Let's add later 😏

jsimonlane:
Hi all, any updates on this?

fantix:
> Hi all, any updates on this?

Sorry, not yet. But we'll put time into the Python binding very soon!

PR #439, opened 2023-06-08
Implement support for vector type

msullivan:
Vectors get decoded into array.array. Encoding supports any list-like
array of numbers, but has an optimized fast path for things like array
and ndarray that avoids needing to box integers.

The test passes when run against the current `pgvector` branch with
edgedb/edgedb#5620 applied to it.

### DIFF
--- a/edgedb/protocol/codecs/codecs.pyx
+++ b/edgedb/protocol/codecs/codecs.pyx
@@ -515,7 +515,7 @@ cdef config_memory_decode(pgproto.CodecContext settings, FRBuffer *buf):
     return datatypes.ConfigMemory(bytes=bytes)
 
 
-DEF PGVECTOR_MAX_DIM = 16000
+DEF PGVECTOR_MAX_DIM = (1 << 16) - 1
 
 
 cdef pgvector_encode_memview(pgproto.CodecContext settings, WriteBuffer buf,
--- a/edgedb/protocol/codecs/codecs.pyx
+++ b/edgedb/protocol/codecs/codecs.pyx
@@ -17,12 +17,15 @@
 #
 
 
+import array
 import decimal
 import uuid
 import datetime
 from edgedb import describe
 from edgedb import enums
 from edgedb.datatypes import datatypes
+
+from libc.string cimport memcpy
 
 
 include "./edb_types.pxi"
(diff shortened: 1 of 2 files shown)


edgedb/protocol/codecs/codecs.pyx
```diff
@@ -510,6 +515,94 @@ cdef config_memory_decode(pgproto.CodecContext settings, FRBuffer *buf):
     return datatypes.ConfigMemory(bytes=bytes)
 
 
+DEF PGVECTOR_MAX_DIM = 16000
```
tailhook, line 518: What's special about this constant that we have to encode this on client?

I mean the datatypes allow encoding 32767 (or even 64k) of values, right? So it's just a matter of time when server (i.e. pgvector extension) might start allowing bigger values. Or is there any extra reserved bit or something?

msullivan: Good point

edgedb/protocol/codecs/codecs.pyx
```diff
@@ -510,6 +515,94 @@ cdef config_memory_decode(pgproto.CodecContext settings, FRBuffer *buf):
...
+        Py_ssize_t i
+
+    objlen = len(obj)
+    if objlen > PGVECTOR_MAX_DIM:
+        raise ValueError('too many elements in vector value')
+
+    buf.write_int32(4 + objlen*4)
+    buf.write_int16(objlen)
+    buf.write_int16(0)
+    for i in range(objlen):
```
1st1, line 535: Can you do `buf.write_cstr(&obj[0], objlen*4)`?

1st1 (approved):

edgedb/protocol/codecs/codecs.pyx
msullivan: Unfortunately we need to convert to network byte order

1st1: Oh god, I see

edgedb/protocol/codecs/codecs.pyx
1st1: > Vectors can have up to 16,000 dimensions.

From the pgvector docs. What will it do if I try to pass an array of 16001 floats? Crash? Trim? Submit and store 16001?

msullivan: Produces an error on the postgres side, which we turn into an InvalidValueError

1st1: OK, sgtm

PR #440, opened 2023-06-09
Fix the vector tests to match the final PR

msullivan:


1st1 (approved):

tests/test_vector.py
```diff
@@ -36,106 +36,96 @@ def setUp(self):
...
             select exists (
-              select sys::ExtensionPackage filter .name = 'vector'
+              select sys::ExtensionPackage filter .name = 'pgvector'
             )
         '''):
             self.skipTest("feature not implemented")
 
         self.client.execute('''
-            create extension vector version '1.0'
+            create extension pgvector version '0.4.0'
```
elprans, line 45: Do we need to be explicit about the version here?  I worry this would be a maintenance burden as soon as bundled versions diverge in different versions of server.

1st1: Oh. Yeah, it should probably be optional? And in the future support `>=` etc

msullivan: Yeah, it is optional. I'll drop it.

PR #450, opened 2023-07-20
Fix tests broken by parser error message changes

msullivan:


elprans:
Fixed in #449

vpetrovykh (approved):
Yeah, let's drop the exact wording for error messages from the binding tests.

PR #451, opened 2023-07-20
Don't depend on exact syntax errors in tests

msullivan:
We have tests for syntax errors in the server, we're just testing the
binding/protocol flow here. (And other places test that messages come
through more generally.)


elprans (approved):

vpetrovykh (approved):
No reason to depend on error message wording.

PR #452, opened 2023-07-21
Add multirange support.

vpetrovykh:
Add multirange codec.

Adjust edgedb.Range and create edgedb.Multirange class as Python represenation of ranges and multiranges.

### DIFF
--- a/edgedb/__init__.py
+++ b/edgedb/__init__.py
@@ -25,7 +25,7 @@ from edgedb.datatypes.datatypes import (
     Tuple, NamedTuple, EnumValue, RelativeDuration, DateDuration, ConfigMemory
 )
 from edgedb.datatypes.datatypes import Set, Object, Array, Link, LinkSet
-from edgedb.datatypes.range import Range
+from edgedb.datatypes.range import Range, Multirange
 
 from .abstract import (
     Executor, AsyncIOExecutor, ReadOnlyExecutor, AsyncIOReadOnlyExecutor,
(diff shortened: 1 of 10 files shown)


edgedb/__init__.py
```diff
@@ -25,7 +25,7 @@
     Tuple, NamedTuple, EnumValue, RelativeDuration, DateDuration, ConfigMemory
 )
 from edgedb.datatypes.datatypes import Set, Object, Array, Link, LinkSet
-from edgedb.datatypes.range import Range
+from edgedb.datatypes.range import Range, Multirange
```
1st1, line 28:
```suggestion
from edgedb.datatypes.range import Range, MultiRange
```

edgedb/datatypes/range.py
```diff
@@ -78,22 +78,31 @@ def is_empty(self) -> bool:
     def __bool__(self):
         return not self.is_empty()
 
-    def __eq__(self, other):
-        if not isinstance(other, Range):
+    def __eq__(self, other) -> bool:
```
msullivan, line 81: I don't think we should have try to have `Range` and `MultiRange` able to compare equal. 
In particular, if we are going to do that, we need to have `__hash__` be equal when the objects are equal, which is not worth dealing with.

vpetrovykh: Good point about hashes equality. I'll drop the equality between Range and MultiRange.

msullivan (approved):

PR #453, opened 2023-07-24
Fix an error in string representation of RelativeDuration.

vpetrovykh:
RelativeDuration was getting the '-' for the seconds component wrong in some cases. Specifically, when fractional seconds were 0, but larger units (hours or seconds) were non zero and negative.

Also add test cases for similar situtation with other duration types.

### DIFF
--- a/edgedb/datatypes/relative_duration.pyx
+++ b/edgedb/datatypes/relative_duration.pyx
@@ -108,7 +108,12 @@ cdef class RelativeDuration:
             buf.append(f'{min}M')
 
         if sec or fsec:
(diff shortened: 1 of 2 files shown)


msullivan (approved):
Well that is some fun.

It might be worth at some point exposing these hours/minutes/second components directly. Seems a shame to go through all of this trouble only to expose them via string formatting

vpetrovykh:
> It might be worth at some point exposing these hours/minutes/second components directly. Seems a shame to go through all of this trouble only to expose them via string formatting

Yeah, we probably want some nice getters or just `@property` for those at some point.

PR #473, opened 2024-01-31
Support passing dicts and namedtuples for namedtuple arguments

### LINKED ISSUE #374: Support way to specify named tuples using names
Currently it seems named tuples can only be specified as tuples, based on order. We probably want to be able to support passing dicts in?

msullivan:
We support dicts (and other mappings), and for named tuples we do
lookup by name instead of by position.

Fixes #374.

### DIFF
--- a/edgedb/protocol/codecs/base.pyx
+++ b/edgedb/protocol/codecs/base.pyx
@@ -271,10 +271,20 @@ cdef class BaseNamedRecordCodec(BaseRecordCodec):
         for i in range(objlen):
             if is_dict:
                 name = datatypes.record_desc_pointer_name(self.descriptor, i)
-                item = obj[name]
+                try:
+                    item = obj[name]
+                except KeyError:
+                    raise ValueError(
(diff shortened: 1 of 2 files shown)


edgedb/protocol/codecs/base.pyx
```diff
@@ -225,6 +227,76 @@ cdef class BaseNamedRecordCodec(BaseRecordCodec):
...
+                f'got {objlen}')
+
+        elem_data = WriteBuffer.new()
+        for i in range(objlen):
+            if is_dict:
+                name = datatypes.record_desc_pointer_name(self.descriptor, i)
+                item = obj[name]
+            elif is_namedtuple:
+                name = datatypes.record_desc_pointer_name(self.descriptor, i)
+                item = getattr(obj, name)
```
elprans, line 277: Probably worth wrapping possible `AttributeError`/`KeyError` to make it nicer.

elprans (approved):

msullivan:
@raddevon We'll probably want some docs about this once we've got a version update out

PR #480, opened 2024-02-10
Add codegen for pgsparse.

vpetrovykh:


vpetrovykh:
I'm a little confused by the test skipping policy w.r.t. codegen. In the rest of the test suite we skip `pgvector` and `pgsparse` tests if the extensions are not detected, but codegen tests simply assume presence of `pgvector` (and `pgsparse` now). Is this something that needs to be updated and made skippable as well?

vpetrovykh:
We'll hold off on adding `pgsparse`, so I'm closing this (and reverting the other `pgsparse` commit).

PR #484, opened 2024-03-12
Add handling of `branch` option.

vpetrovykh:
The new `branch` option can be used instead of the old `database` option. These are mutually exclusive and will generally produce an error if used at the same time (except when one appears in settings that completely override other sources of options).

The branch option can be passed in the following manner:
- as a kwarg `branch` to the client
- as `EDGEDB_BRANCH` environment variable
- as a `branch` field of the credentials
- as a `branch` query parameter in DSN

This binding will accept either `branch` or `database` option and will normalize the result as `database` when connecting to the EdgeDB server (for compatibility with older servers during the period of deprecation of `database` term).

### DIFF
--- a/edgedb/base_client.py
+++ b/edgedb/base_client.py
@@ -118,6 +118,10 @@ class BaseConnection(metaclass=abc.ABCMeta):
     def dbname(self) -> str:
         return self._params.database
 
+    @property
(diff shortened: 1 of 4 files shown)


fantix (approved):
LGTM!

PR #487, opened 2024-04-05
Make conflict detection between branch and database less strict.

vpetrovykh:
It is an error to specify `database` and `branch` at the same time on the same configuration level.
It is OK to override previous configuration with either `database` or `branch`.
It is OK for the credentials file to contain both `database` and `branch` fields in order to facilitate transitioning from old style to the new style while maintaining backwards compatibility. The values must agree, though.

### DIFF
--- a/edgedb/con_utils.py
+++ b/edgedb/con_utils.py
@@ -182,8 +182,8 @@ class ResolvedConnectConfig:
     _port_source = None
 
     # We keep track of database and branch separately, because we want to make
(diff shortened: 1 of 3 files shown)


aljazerzen (approved):

aljazerzen:
I've implemented the same behavior in edgedb-rust.

When writing credentials.json, I've even implemented writing *both* branch and database, so it is backward compatible. This is useful if credentials.json is written by CLI on the new version and then old version of bindings reads it.

PR #490, opened 2024-05-01
Expose embeddings API

fantix:
This PR adds `EdgeDBAI.generate_custom_embeddings()` function to allow users to generate embedding vectors without using the AI index in EdgeDB with custom input text.

I'll create another PR to add docs.


edgedb/ai/core.py
```diff
@@ -112,23 +125,26 @@ def query_rag(
...
-                query=message,
+                message=message,
                 stream=True,
             ).to_httpx_request(),
         ) as event_source:
             event_source.response.raise_for_status()
             for sse in event_source.iter_sse():
                 yield sse.data
 
+    def generate_custom_embeddings(self, *inputs: str, model: str):
```
scotttrinh, line 141: working on the JS version of this but my instinct was to name this `generateEmbeddings`. Is there a significance to the "custom" here? Does it help to denote that these are not the automatically indexed ones?

fantix: Yeah, that was my intention, but I'm not sure this makes sense (like in English).

scotttrinh: I wonder if `custom` communicates enough to pay for the cost. `search` would be even more specific and point at the intended use case, but maybe that's too restrictive since you might use it outside of `ext::ai::search`? Does `generate_embeddings` make it seem too much like you're triggering something rather than doing this one-off embedding generation?

The only reason I ask is that having "custom" here would make me as a developer want to know more about what "custom" means and what other non-custom methods their might be. Maybe that's a good thing and worth adding here, but it feels a little like an unnecessary mental speed bump.

fantix: > Does `generate_embeddings` make it seem too much like you're triggering something rather than doing this one-off embedding generation?

Yeah, I had the same struggle. I thought about `retrieve_embeddings()`, which is just worse.

> Maybe that's a good thing and worth adding here, but it feels a little like an unnecessary mental speed bump.

Right, I'll also change it to `generate_embeddings()` and add an explanation in the docs.

edgedb/ai/core.py
```diff
@@ -156,19 +168,22 @@ async def query_rag(
...
             event_source.response.raise_for_status()
             async for sse in event_source.aiter_sse():
                 yield sse.data
+
+    async def generate_embeddings(self, *inputs: str, model: str):
+        resp = await self.client.post(
+            "/embeddings", json={"input": inputs, "model": model}
+        )
+        resp.raise_for_status()
+        return resp.json()
```
elprans, line 189: This should return a float list like `edgedb-js` does

edgedb/ai/core.py
```diff
@@ -155,20 +169,25 @@ async def query_rag(
 
     async def stream_rag(
         self, message: str, context: typing.Optional[types.QueryContext] = None
-    ):
-        if context is None:
-            context = self.context
+    ) -> typing.Iterator[str]:
```
fantix, line 172: This returned `str` is a JSON string like:

```json
{"type": "content_block_delta","index":0,"delta":{"type": "text_delta", "text": " blocking"}}
```

Issue #486, opened 2024-04-03
BlockingIOConnection doesn't support IPV6


cmr-a:
**Describe the bug**
BlockingIOConnection throws when you try to connect to an IPv6 address

**Reproduction**
```python
import edgedb

client = edgedb.create_client(dsn="edgedb://edgedb:edgedb@[::1]:5656/edgedb")
client.ensure_connected()
```

Produces:
```
gaierror                                  Traceback (most recent call last)
File [ ~/venv/site-packages/edgedb/blocking_client.py:52], in BlockingIOConnection.connect_addr(self, addr, timeout)
     [57] try:
---> [58]     sock.connect(addr)
     [60]     if not isinstance(addr, str):

gaierror: [Errno -9] Address family for hostname not supported

...
```

The exception is thrown here: https://github.com/edgedb/edgedb-python/blob/c666a6f09921e683278762a9c9fe9916cbb0348e/edgedb/blocking_client.py#L58

**Expected behavior**
Connection established / connection refused

**Versions (please complete the following information):**

- OS: Linux
- EdgeDB version: N/A
- EdgeDB CLI version: 4.0.2
- `edgedb-python` version: 1.9.0
- Python version: 3.11.8

**Additional context**
Other IPv6 addresses produce the same error, nothing special about [::1].

No problems with asyncio_client.

I was able to fix blocking_client.py with:
```python
        if isinstance(addr, str):
            # UNIX socket
            sock = socket.socket(socket.AF_UNIX)
        else:
            if ":" in addr[0]:
                sock = socket.socket(socket.AF_INET6)
            else:
                sock = socket.socket(socket.AF_INET)
```

I don't know if this is the right way to fix it, but it allowed the connection to be established.

zachary822:
This new branch should fix the issue. Can you give it a try?

cmr-a:
It works, thanks!

PR #491, opened 2024-05-04
fix ipv6 connection issues

zachary822:
Addresses #486

Uses [`socket.getaddrinfo`](https://docs.python.org/3/library/socket.html#socket.getaddrinfo) to get address family from provided host and port for the blocking client. This addresses the previous issue that only ipv4 clients were allowed.


cmr-a:
I tested this. I was able to connect to a remote edgedb instance over ipv6 with these changes. Thanks for the PR!

fantix:
Thanks for the PR! However, I think we should try to connect to all resolved addresses and use the first one that succeeded, in the same way as asyncio does. I submitted #499 (with your commit) to also fix IPv6.

PR #501, opened 2024-05-31
Use result of pydantic_dataclass, will silence linters

AdrienPensart:
This fix will silence some linters warning about unused results.


fantix (approved):
Thanks!

PR #522, opened 2024-09-27
Add more pgvector types.

vpetrovykh:
Add support for `sparsevec` and `halfvec`.
`halfvec` is handled as a float array (converted from float32 to float16 or vice versa as needed).
`sparsevec` is handled as a dict with non-zero values and "dimensions" specified in it.

### DIFF
--- a/edgedb/protocol/codecs/codecs.pyx
+++ b/edgedb/protocol/codecs/codecs.pyx
@@ -67,6 +67,12 @@ cdef BaseCodec NULL_CODEC = NullCodec.__new__(NullCodec)
 cdef BaseCodec EMPTY_TUPLE_CODEC = EmptyTupleCodec.__new__(EmptyTupleCodec)
 
 
+# Needed to perform bit manipulations to convert from float16 to float32
+cdef union IntFloatSwap:
+    uint32_t i
+    float f
+
+
 cdef class CodecsRegistry:
 
     def __init__(self, *, cache_size=1000):
@@ -716,7 +722,6 @@ DEF PGVECTOR_MAX_DIM = (1 << 16) - 1
 cdef pgvector_encode_memview(pgproto.CodecContext settings, WriteBuffer buf,
                              float[:] obj):
     cdef:
-        float item
         Py_ssize_t objlen
         Py_ssize_t i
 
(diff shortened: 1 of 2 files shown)


vpetrovykh:
The `halfvec` codecs need more robust conversion for float16. This seems to outline a good algo that I can implement here: http://www.fox-toolkit.org/ftp/fasthalffloatconversion.pdf.

Sometimes `_Float16` might be available via `#include <float.h>`, which should take care of casting between `float`/`double` and `float16`, but it would still need a fall-back.

edgedb/protocol/codecs/codecs.pyx
```diff
@@ -798,6 +796,264 @@ cdef pgvector_decode(pgproto.CodecContext settings, FRBuffer *buf):
...
+        e -= 0x00800000         # Decrement exponent (1<<23)
+        m <<= 1                 # Shift mantissa
+
+    m &= ~0x00800000            # Clear leading 1 bit
+    e += 0x38800000             # Adjust bias ((127-14)<<23)
+
+    return m | e
+
+
+cdef generate_tables():
```
msullivan, line 832: Could you put a comment in here (and the final commit message) with a cite to the algorithm you've implemented?

edgedb/protocol/codecs/codecs.pyx
```diff
@@ -798,6 +796,264 @@ cdef pgvector_decode(pgproto.CodecContext settings, FRBuffer *buf):
...
+    return val
+
+
+cdef pgvector_sv_encode(pgproto.CodecContext settings, WriteBuffer buf,
+                        object obj):
+    cdef:
+        int32_t dim
+        Py_ssize_t nnz
+        # int32_t key
+        # float val
```
msullivan, line 985: Dead

msullivan (approved):

edgedb/protocol/codecs/codecs.pyx
vpetrovykh: Comment added

edgedb/protocol/codecs/codecs.pyx
vpetrovykh: Cleaned up

PR #526, opened 2024-11-04
Support Python 3.13

msullivan:


edgedb/protocol/protocol.pxd
```diff
@@ -41,10 +41,10 @@ ctypedef object (*decode_row_method)(BaseCodec, FRBuffer *buf)
...
 
 cpdef enum OutputFormat:
-    BINARY = b'b'
-    JSON = b'j'
-    JSON_ELEMENTS = b'J'
-    NONE = b'n'
+    BINARY = 98  # b'b'
+    JSON = 106  # b'j'
+    JSON_ELEMENTS = 74  # b'J'
+    NONE = 110  # b'n'
```
1st1, line 47: weird.

1st1 (approved):

setup.py
```diff
@@ -38,7 +38,7 @@
 from setuptools.command import sdist as setuptools_sdist
 
 
-CYTHON_DEPENDENCY = 'Cython(>=0.29.24,<0.30.0)'
+CYTHON_DEPENDENCY = 'Cython(>=3.0.11,<3.1.0)'
```
elprans, line 41: Please leave the lower bound on 0.29.x before every other Cython package is upgraded (most importantly, server).  This is for metapkg reasons.

msullivan: 0.29.x doesn't support 3.13, AFAICT

PR #532, opened 2024-11-07
Support GEL_ env vars and gel.toml

msullivan:


edgedb/con_utils.py
```diff
@@ -1169,18 +1133,19 @@ def find_edgedb_project_dir():
     dev = os.stat(dir).st_dev
 
     while True:
-        toml = os.path.join(dir, 'edgedb.toml')
-        if not os.path.isfile(toml):
+        gel_toml = os.path.join(dir, 'gel.toml')
```
mmastrac, line 1136: Do we want to have the same logic where we check the file contents between the two here?

mmastrac (approved):
LGTM, one small question

edgedb/con_utils.py
msullivan: I don't think so, since we don't currently read the file at all?

PR #536, opened 2024-11-14
Gel sqla

vpetrovykh:
More tests coming...

### DIFF
--- a/gel/_testbase.py
+++ b/gel/_testbase.py
@@ -21,6 +21,7 @@ import asyncio
 import atexit
 import contextlib
 import functools
+import importlib.util
 import inspect
 import json
 import logging
@@ -35,6 +36,8 @@ import unittest
 import gel
 from gel import asyncio_client
 from gel import blocking_client
+from gel.orm.introspection import get_schema_json
+from gel.orm.sqla import ModelGenerator
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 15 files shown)


edgedb/gelalchemy/models.py
```diff
@@ -0,0 +1,324 @@
...
+class ModelGenerator(object):
+    INDENT = ' ' * 4
+
+    def __init__(self, *, out=None):
+        if out is not None:
+            self.out = out
+        else:
+            self.out = sys.stdout
+
+        self._indet_level = 0
```
msullivan, line 68: `_indent_level`

msullivan:
Could you post some generated code?

edgedb/_testbase.py
```diff
@@ -594,6 +599,34 @@ def adapt_call(cls, result):
...
+        super().setUpClass()
+
+        class_set_up = os.environ.get('EDGEDB_TEST_CASES_SET_UP')
+        if not class_set_up:
+            # Now that the DB is setup, generate the SQLAlchemy models from it
+            spec = get_schema_json(cls.client)
+
+            with open(cls.SQLAMODELS, 'w') as f:
+                gen = ModelGenerator(out=f)
+                gen.render_models(spec)
```
msullivan, line 617: We should write the target module out into a temporary directory that we add to `sys.path`, instead of writing to files in the source tree

edgedb/gelalchemy/models.py
```diff
@@ -0,0 +1,390 @@
...
+                self.indent()
+                self.write(f'{lname!r}, Uuid(), ForeignKey("{target}.id"),')
+                self.write(f'primary_key=True, nullable=False,')
+                self.dedent()
+                self.write(')')
+
+                if lname == 'source':
+                    bklink = source_link
+                else:
+                    bklink = f'backlink_via_{source_link}'
```
msullivan, line 147: What is there are multiple incoming links with that name?

edgedb/gelalchemy/models.py
vpetrovykh: Fixed

edgedb/_testbase.py
vpetrovykh: It's a temp directory now.

edgedb/gelalchemy/models.py
vpetrovykh: I don't really have a handler for backlink name collisions. I'm tempted to just start adding numbers to them because adding anything longer to the names (referring to the object the backlink is from) is probably too much for a name.

vpetrovykh: Fixed

gel/orm/introspection.py
```diff
@@ -197,6 +199,32 @@ def _process_links(types, modules):
...
+                for link in bklinks:
+                    origsrc = get_sql_name(link['target']['name'])
+                    lname = link['name']
+                    link['name'] = f'{lname}_from_{origsrc}'
+                    # Also update the original source of the link with the
+                    # special backlink name.
+                    source = type_map[link['target']['name']]
+                    fwname = lname.replace('backlink_via_', '', 1)
+                    link['fwname'] = fwname
+                    source['backlink_renames'][fwname] = link['name']
```
msullivan, line 226: Could we *always* generate the disambiguated backlink, and then only generate the short name when it is unambiguous, or does that not work?

vpetrovykh: We mostly expect not too many collisions so I tend to treat that as a the default. Backlink collisions appear when everything is linked to everything, which is not necessarily a common schema since that creates a lot of links to maintain.
Also, detecting unambiguous backlinks is a little more annoying that the other way around. Collisions are shown by identical names, but unambiguous long names would have to be checked by looking up their corresponding forward link name, seems more roundabout.

gel/_testbase.py
```diff
@@ -623,6 +630,54 @@ def adapt_call(cls, result):
...
+    DEFAULT_MODULE = 'default'
+
+    @classmethod
+    def setUpClass(cls):
+        # SQLAlchemy relies on psycopg2 to connect to Postgres and thus we
+        # need it to run tests. Unfortunately not all test environemnts might
+        # have psycopg2 installed, as long as we run this in the test
+        # environments that have this, it is fine since we're not expecting
+        # different functionality based on flavours of psycopg2.
+        if importlib.util.find_spec("psycopg2") is None:
```
msullivan, line 644: I think it would be cleaner to do 
```
try:
    import psychopg2
except ImportError:
    # handle it
```

msullivan (approved):

gel/_testbase.py
vpetrovykh: I wanted to avoid the import since I don't even _need_ anything from `psycopg2` directly, not using it anywhere myself. Didn't want to leave the impression that I needed the import myself and if it's not used, maybe it's OK to "remove this dependency check" making it look like an outdated comment or dead code. `importlib` seemed like a more overt and unambiguous check.

PR #541, opened 2024-11-16
Add package alias with new name

msullivan:
Add a `gel` alias that does `import *` from `edgedb`.

Note that this will *not* provide any submodules.
Doing `from gel import errors` won't work.

But we also never documented anything internal, and it will all keep
working with an `edgedb` import, so nothing will break.

If we feel strongly about providing all the internal modules under the
`gel` name also, we have options, but as far as I can tell they are
all annoying in some way:
 - Generate a parallel tree of modules under `gel` that all do `import *`
   from an appropriate `edgedb` module. This is kind of ugly and also
   we need to regenerate when anything changes.
 - Something involving hooking the python import system. I think this
   would work but also I think involves modifying global python interpreter
   state when `gel` is imported, which is dodgy.

I'd probably go with option 1 if we care, but not caring seems fine?

### DIFF
--- a/gel/__init__.py
+++ b/gel/__init__.py
@@ -1,0 +1,22 @@
+#
+# This source file is part of the EdgeDB open source project.
+#
+# Copyright 2016-present MagicStack Inc. and the EdgeDB authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


1st1 (changes requested):
I think we should just rename and tell to people to start using "gel" if they want new stuff.

PR #544, opened 2024-11-19
Rename main module from edgedb to gel

msullivan:
Put some annoying shims in to keep from breaking edgedb package,
though.


msullivan:
Since `_version.py` got modified, I disabled the release action to keep it from being mad that the release was malformed. We'll reenable it once this is merged.

fantix (approved):
:+1:

PR #555, opened 2024-12-03
API fix: with_query_tag()/without_query_tag()

fantix:


gel/options.py
```diff
@@ -413,17 +414,25 @@ def without_globals(self, *global_names):
         )
         return result
 
-    def with_annotation(self, name: str, value: str):
+    def with_query_tag(self, tag: str):
+        for prefix in ["edgedb/", "gel/"]:
+            if tag.startswith(prefix):
+                raise errors.InvalidArgumentError(f"reserved tag: {prefix}*")
```
1st1, line 420: Maybe just a warning? The server should ideally just silently ignore them

fantix: I thought the same, but we will send e.g. `gel/repl` tags from the CLI or `gel/ui` from the UI/JS client, while the server has no knowledge about the peer identity; the server could also receive `myapp/financial-queries` equally, so it just records the tags as received without ignoring. Unless you meant to reserve `gel/*` for only server-internal queries like in some table GC tasks (instead of queries over the binary protocol), adding an error in client bindings is the last chance we can enforce this tag namespace.

gel/options.py
```diff
@@ -413,17 +414,25 @@ def without_globals(self, *global_names):
...
+        if len(tag) > 128:
+            raise errors.InvalidArgumentError(
+                "tag too long (> 128 characters)"
+            )
```
msullivan, line 424: Do we check this server-side also?

msullivan (approved):

gel/options.py
fantix: Yes, we do!

msullivan:
Should we merge this?

PR #556, opened 2024-12-09
Add a basic Django model generator.

vpetrovykh:
Due to limitations of Django the schema must only have the `default` module. Also all link properties and multi properties are forbidden since they produce models without a unique `id` (or other primary key) column. Django cannot properly handle updates to that.

Added tests for the generated Django models.

### DIFF
--- a/gel/orm/introspection.py
+++ b/gel/orm/introspection.py
@@ -1,6 +1,7 @@
 import json
 import re
 import collections
+import textwrap
 
 
 INTRO_QUERY = '''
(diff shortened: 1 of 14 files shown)


gel/orm/introspection.py
```diff
@@ -232,3 +244,29 @@ def _process_links(types, modules):
...
+
+    def indent(self):
+        self._indent_level += 1
+
+    def dedent(self):
+        if self._indent_level > 0:
+            self._indent_level -= 1
+
+    def reset_indent(self):
+        self._indent_level -= 0
```
msullivan, line 266: I'm guessing you want `=` here instead

msullivan (approved):
Could you post the generated output for the test in a gist or something, so we can look at it easily?

gel/orm/introspection.py
vpetrovykh: Yeah, I fixed it to `=` :)

vpetrovykh:
Here are the models generated from the schema I used for Django tests:
https://gist.github.com/vpetrovykh/ef93630efdb3ca13cda2b144c8825c7f

PR #560, opened 2024-12-12
Implement gel.Record type

1st1:
It's returned from `client.query_sql()` and represents SQL rows.

Currently it supports:

* integer indexing - fetch columns by position
* string indexing - fetch columns by name
* as_dict() method - render record as a dict

ToDo:

* add __iter__

### DIFF
--- a/gel/datatypes/object.c
+++ b/gel/datatypes/object.c
@@ -124,25 +124,6 @@ EdgeObject_GetItem(PyObject *ob, Py_ssize_t i)
 }
 
 
-PyObject *
-EdgeObject_GetID(PyObject *ob)
-{
-    assert(EdgeObject_Check(ob));
-    EdgeObject *o = (EdgeObject *)ob;
-    Py_ssize_t i = EdgeRecordDesc_IDPos(o->desc);
-    if (i < 0) {
-        Py_RETURN_NONE;
-    }
-    if (i >= Py_SIZE(o)) {
-        PyErr_BadInternalCall();
-        return NULL;
-    }
-    PyObject *el = EdgeObject_GET_ITEM(o, i);
-    Py_INCREF(el);
-    return el;
-}
-
-
 static void
 object_dealloc(EdgeObject *o)
 {
@@ -284,7 +265,7 @@ object_repr(EdgeObject *o)
 
     if (_EdgeGeneric_RenderItems(&writer,
                                  (PyObject *)o, o->desc,
-                                 o->ob_item, Py_SIZE(o), 1, 0) < 0)
+                                 o->ob_item, Py_SIZE(o), 1, 1, 0) < 0)
     {
         goto error;
     }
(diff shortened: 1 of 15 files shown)


gel/datatypes/object.c
```diff
@@ -284,7 +265,7 @@ object_repr(EdgeObject *o)
 
     if (_EdgeGeneric_RenderItems(&writer,
                                  (PyObject *)o, o->desc,
-                                 o->ob_item, Py_SIZE(o), 1, 0) < 0)
+                                 o->ob_item, Py_SIZE(o), 1, 1, 0) < 0)
```
msullivan, line 268: You don't really need to fix this now, but this sort of list of boolean arguments ought to either have comments saying what each of the argument names are or should use #defines/enums or flag bits or something. `1, 1, 0` is inscrutable.

This is an existing problem, though, so it's fine to skip fixing it now.

gel/datatypes/record.c
```diff
@@ -0,0 +1,294 @@
...
+
+
+static PyObject *
+record_repr(EdgeRecord *o)
+{
+    _PyUnicodeWriter writer;
+    _PyUnicodeWriter_Init(&writer);
+    writer.overallocate = 1;
+
+    if (_PyUnicodeWriter_WriteASCIIString(&writer, "Row{", 4) < 0) {
```
msullivan, line 203: Should this be `Record{`?

tests/test_async_query.py
```diff
@@ -1151,3 +1151,13 @@ async def test_transaction_state(self):
...
+
+    # async def test_async_query_sql_01(self):
+    #     res = await self.client.query_sql("SELECT 1")
+    #     self.assertEqual(res[0].as_dict(), {'col~1': 1})
+
+    #     res = await self.client.query_sql("SELECT 1 as aa")
+    #     self.assertEqual(res[0].as_dict(), {'aa': 1})
+
+    #     res = await self.client.query_sql("SELECT FROM generate_series(0, 1)")
+    #     self.assertEqual(res[0].as_dict(), {})
```
msullivan, line 1163: Use `xfail` or `skip` instead of commenting out

msullivan (approved):
Looks reasonable. Kind of a lot of duplication, but you said that merging it would be worse?

1st1:
@msullivan 

> Kind of a lot of duplication

What's duplicated? Besides the Record codec looking like a stripped down Object codec?

tests/test_async_query.py
1st1: sure

gel/datatypes/record.c
1st1: fixed

gel/datatypes/object.c
1st1: Fixed in [Apply better style for passing flags](https://github.com/edgedb/edgedb-python/pull/560/commits/c62afea58c1bcb46759038e2354d1437ea20dd8b)

msullivan: Much nicer

PR #562, opened 2024-12-13
A bunch of tests that were mismatched between sycn/async.

vpetrovykh:
SyncQueryTestCase base type was accidentally used with async test methods. Now thye are all sync.

### DIFF
--- a/tests/test_async_query.py
+++ b/tests/test_async_query.py
@@ -1152,8 +1152,12 @@ class TestAsyncQuery(tb.AsyncQueryTestCase):
                     INSERT test::Tmp { id := <uuid>$0, tmp := '' }
                 ''', uuid.uuid4())
 
-    @unittest.expectedFailure
     async def test_async_query_sql_01(self):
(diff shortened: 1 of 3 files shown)


tests/test_async_query.py
```diff
@@ -1152,8 +1152,12 @@ async def test_transaction_state(self):
                     INSERT test::Tmp { id := <uuid>$0, tmp := '' }
                 ''', uuid.uuid4())
 
-    @unittest.expectedFailure
     async def test_async_query_sql_01(self):
+        if await self.client.query_required_single('''
+            select sys::get_version().major < 6
+        '''):
+            self.skipTest("Buggy in versions earlier than 6.0")
```
msullivan, line 1159: What happens? Probably we want some useful error mode there.

msullivan (approved):

tests/test_async_query.py
msullivan: Like, edgedb-python should refuse to do query_sql if it isn't on proto v3

PR #564, opened 2025-01-21
Sqlmodel

vpetrovykh:
There are two commits. The first one is mostly tweaks based on the feedback about the ORM schema generators. The second one adds SQLModel schema generator (that's very similar to SQLAlchemy).


msullivan (approved):
Is there anything in specific you want me to look closely at? I mostly skimmed this, since it is very similar to sqlalchemy

PR #575, opened 2025-02-03
Update

vpetrovykh:
Some more updates to ORM generators.

Mainly this update is about changing the backlink naming scheme to `_<link>_<Type>` format across all of the Python ORM generators.

Add some tests.


aljazerzen:
Please don't tag the whole team as reviewers.

msullivan (approved):

PR #576, opened 2025-02-05
Update python docs

diksipav:


docs/usage.rst
```diff
@@ -122,7 +122,7 @@ Below is an example of a web API server running `aiohttp
             database='my_service',
             user='my_service')
         # Configure service routes
-        app.router.add_route('GET', '/user/{name:\w+}', handle)
```
diksipav, line 125: python \interprets "\w" as an escape sequence, so I removed this regex part
but ofc, can also use the raw string or escape the backslash "\\w+"

scotttrinh: Oh, I think your change makes sense and feels idiomatic.

scotttrinh (approved):

PR #577, opened 2025-02-13
Retry calls to `execute` as well as to `query`

msullivan:
In the distant past, execute was quite different than query (execute
couldn't take arguments and query could not include multiple
commands), so it maybe made sense to have different retry behaviors.

Now, execute and query are basically identical except for whether data
is returned, and so I don't think there is any reason why they
shouldn't both do retries.

(I modeled the implementation after how I did this in the edgedb test
suite's hacked up client: edgedb/edgedb#8249)


fantix (approved):
LGTM!

PR #586, opened 2025-02-26
Add auth extension handling

fantix:
This adds the `gel.auth` module as the general binding of the Gel auth extension.

- [ ] Add test


gel/auth/email_password.py
```diff
@@ -0,0 +1,550 @@
...
+    ) -> EmailVerificationResponse:
+        logger.info("verifying email")
+        verify_response = await self._http_request(
+            "POST",
+            "/verify",
+            json={
+                "verification_token": verification_token,
+                "provider": self.provider,
+            },
+        )
```
fantix, line 330: Is it expected that the verification will proceed (even successfully) with or without a valid PKCE verifier?

scotttrinh: Yes, that's right. PKCE just protects the actual auth token exchange, so you can verify an email from any user agent, but you will not be able to then parlay that into an authentication unless you have the PKCE verifier.

gel/ai/core.py
```diff
@@ -45,25 +45,26 @@ class BaseRAGClient:
...
     def __init__(
         self,
-        client: typing.Union[gel.Client, gel.AsyncIOClient],
+        info: gel.ConnectionInfo,
         options: types.RAGOptions,
         **kwargs,
     ):
-        pool = client._impl
-        host, port = pool._working_addr
-        params = pool._working_params
```
scotttrinh, line 54: Not a blocker, but I think this diff could be a little smaller and focus if you kept this alias like:

```py
        params = info.params
```

I think that would reduce the downstream changes in the rest of this `__init__` def?

gel/auth/pkce.py
```diff
@@ -0,0 +1,125 @@
...
+            token_response.status_code,
+            token_response.text,
+        )
+        token_response.raise_for_status()
+        token_json = token_response.json()
+        args = {
+            field.name: token_json[field.name]
+            for field in dataclasses.fields(token_data.TokenData)
+        }
+        return token_data.TokenData(**args)
```
scotttrinh, line 89: Oh, nice trick, I'm going to stash this one away for myself.

gel/auth/email_password.py
```diff
@@ -0,0 +1,539 @@
...
+            base_url = httpx.URL(
+                scheme=scheme,
+                host=connection_info.host,
+                port=connection_info.port,
+            )
+            kwargs["base_url"] = base_url.join(
+                f"branch/{params.branch}/ext/auth"
+            )
+        if "verify" not in kwargs:
+            kwargs["verify"] = connection_info.params.make_ssl_ctx()
```
scotttrinh, line 167: What does this part do?

scotttrinh: Oh, nevermind, i was getting this confused with email verification, but this is TLS stuff for the httpx client, got it.

PR #587, opened 2025-02-27
Enable multiple modules for SQLModel reflection.

vpetrovykh:


msullivan (approved):

msullivan (commented):
What was preventing it from working before?

vpetrovykh:
> What was preventing it from working before?

Looking back at it I think it was some other bug mostly related to link tables for multi links that made it look like the schema name was incorrectly quoted in the query.

PR #594, opened 2025-03-17
Add encode and decode for array of array.

dnwpark:


fantix (approved):

fantix:
Let's fix so that it doesn't fail at least on the stable servers - if the nightly servers are WIP to get a green badge.

PR #600, opened 2025-03-22
Drop Python 3.8 support

elprans:
Python 3.8 is EOL


fantix (approved):
This is good; I'll create another PR to drop 3.8 codegen logic.

PR #609, opened 2025-04-14
Implement PreferRepeatableRead

msullivan:
COMMENTS NEEDED: We may want to reconsider the interface the user uses
to specify it!
Currently it works by setting default_transaction_isolation to the
enum value PreferRepeatableRead.

If that is set, the parse and execute paths will strip it from the
state that is sent, and then potentially inject RepeatableRead as the
value when doing execute, if it is safe.

### DIFF
--- a/gel/abstract.py
+++ b/gel/abstract.py
@@ -69,6 +69,7 @@ class QueryContext(typing.NamedTuple):
     state: typing.Optional[options.State]
     warning_handler: options.WarningHandler
     annotations: typing.Dict[str, str]
+    transaction_options: typing.Optional[options.TransactionOptions]
 
     def lower(
         self, *, allow_capabilities: enums.Capability
@@ -86,6 +87,7 @@ class QueryContext(typing.NamedTuple):
             allow_capabilities=allow_capabilities,
             state=self.state.as_dict() if self.state else None,
             annotations=self.annotations,
+            transaction_options=self.transaction_options,
         )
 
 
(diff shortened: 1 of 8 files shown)


1st1:
@msullivan I started reviewing this but quickly realized that I probably should implement it in JS myself to better understand the scope. Then I'll be able to advise on this PR. I'll get to it today.

elprans:
> COMMENTS NEEDED: We may want to reconsider the interface the user uses to specify it! Currently it works by setting default_transaction_isolation to the enum value PreferRepeatableRead.

One danger of adding the enum value is possible confusion that it's a setting that the _server_ supports (but this is minor since we'll add this server-side in 7).

Interface-wise, we should make `with_transaction_options` apply to implicit transactions as well.

elprans (approved):

msullivan:
> Interface-wise, we should make `with_transaction_options` apply to implicit transactions as well.

I'm not sold on that exactly. What would be the interaction with explicitly setting the config values?

elprans:
> > Interface-wise, we should make `with_transaction_options` apply to implicit transactions as well.
> 
> I'm not sold on that exactly. What would be the interaction with explicitly setting the config values?

`with_transaction_options` will override the config settings, which makes sense because it's an even more "granular" scope than a session.

msullivan:
> > > Interface-wise, we should make `with_transaction_options` apply to implicit transactions as well.
> > 
> > 
> > I'm not sold on that exactly. What would be the interaction with explicitly setting the config values?
> 
> `with_transaction_options` will override the config settings, which makes sense because it's an even more "granular" scope than a session.

That's plausible, but we'll have to make them default to `None` instead of actual values, but that's doable

msullivan:
Alright I think this is done, now?

elprans (approved):

msullivan:
darn it, I accidentally hit enter while still editing the commit message

PR #612, opened 2025-04-22
Make TransactionOptions default to optional

msullivan:
Visible behavior should be the same for now. Later, once we support
putting the TransactionOptions values into config, it will be
important to distinguish whether they are actually set.

Extracted from #609.


elprans (approved):

scotttrinh (approved):
Interesting. OK, I can match this in the JS driver as well.

PR #616, opened 2025-04-24
Support PreferRepeatableRead on explicit transactions by retrying

msullivan:
The method for detecting the need to retry is kind of hacky.

We cache using the filename and line number of the caller.

### DIFF
--- a/gel/base_client.py
+++ b/gel/base_client.py
@@ -415,6 +415,7 @@ class BasePoolImpl(abc.ABC):
         "_connect_args",
         "_codecs_registry",
         "_query_cache",
+        "_tx_needs_serializable_cache",
         "_connection_factory",
         "_queue",
         "_user_max_concurrency",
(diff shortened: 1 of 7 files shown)


scotttrinh (approved):
Makes sense to me, and matches what Yury and I did in the JS (minus the caching). 👍

1st1:
I don't think we should implement the cache. Anyone who inherits from our client will have a broken caching. Like this is what Credal is doing with gel-js.

msullivan:
> I don't think we should implement the cache. Anyone who inherits from our client will have a broken caching. Like this is what Credal is doing with gel-js.

I don't fully follow?

I did put an escape hatch in the cache for dealing with cases where the client always has an extra wrapper function, or similar. We can also tweak the cache behavior so that it doesn't *always* default to serializable forever once it has been needed once; we could flip the state back.

1st1:
> I did put an escape hatch in the cache for dealing with cases where the client always has an extra wrapper function, or similar.

Yeah, I saw that and the explaining comment, I'm just really on the fence here thinking, hey -- it's way too much magic, and then I'm flipping back thinking -- hey, it can save a bunch of roundtrips and should be done. Anyway, let's go with the cache, let's do it. Approving the PR.

1st1 (approved):

PR #623, opened 2025-05-11
Add query batching API

fantix:
```python
import asyncio
import gel

client = gel.create_async_client()

async def main():
    async for b in client._batch():
        async with b:
            await b.send_query_single("select 42 + <int32>$0", 2)
            await b.send_query_single("select 42 * <float32>$0", 2)
            x, y = await b.wait()
            assert x == 44
            assert y == 84.0
            print(x, y)

asyncio.run(main())
```

For manual transaction isolation, use `client.with_transaction_options()`.

- [x] Test and fix
- [ ] Send messages immediately in `send_query_*()`
- [x] Add test
- [x] Add backpressure
- [x] Add retrying transaction
- [x] Add blocking API
- [x] Retry on compatible new codecs without cache

### DIFF
--- a/gel/asyncio_client.py
+++ b/gel/asyncio_client.py
@@ -367,6 +367,7 @@ class AsyncIOBatchIteration(transaction.BaseTransaction):
         return self
 
     async def __aexit__(self, extype, ex, tb):
+        await self.wait()
         with self._exclusive():
             self._managed = False
             return await self._exit(extype, ex)
--- a/gel/asyncio_client.py
+++ b/gel/asyncio_client.py
@@ -349,55 +349,48 @@ class AsyncIORetry(transaction.BaseRetry):
         return iteration
 
 
-class AsyncIOBatch:
-    def __init__(
-        self, client: AsyncIOClient, optimistic_isolation: bool
-    ) -> None:
-        self._client = client
-        self._optimistic_isolation = optimistic_isolation
+class AsyncIOBatchIteration(transaction.BaseTransaction):
+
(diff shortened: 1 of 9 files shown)


gel/asyncio_client.py
```diff
@@ -346,6 +349,123 @@ async def __anext__(self):
...
+
+    async def __aenter__(self):
+        if self._managed:
+            raise errors.InterfaceError(
+                'cannot enter context: already in an `async with` block')
+        self._managed = True
+        return self
+
+    async def __aexit__(self, extype, ex, tb):
+        with self._exclusive():
```
1st1, line 370: exit should call `await self.wait()` if there are some batched but never waited for operations, right?

fantix: oh yes, good catch!

1st1: please add/generate some tests!

gel/protocol/protocol.pyx
```diff
@@ -553,6 +645,137 @@ cdef class SansIOProtocol:
...
+                                exc.tags = exc.tags.union(
+                                    {errors.SHOULD_RETRY}
+                                )
+
+                                # If we may retry this batch, invalidate codec
+                                # cache for all queries except the current one
+                                # because we just updated its input codec
+                                for c in ctxs:
+                                    if c is not ctx:
+                                        c.invalidate_cache()
```
1st1, line 753: We need to exit the loop or re-raise? Not sure I see what happens next after invalidation here

1st1 (approved):

PR #624, opened 2025-05-13
Retry on input type mismatches if compatible

fantix:
If the cached codec mismatches the server schema, we shall allow retrying if the new codec encodes the user arguments successfully.


1st1:
@jaclarke @scotttrinh we should do this for gel-js too

msullivan (approved):

PR #625, opened 2025-05-26
Allow specifying `allow_user_specified_id` for codegen

Lur1an:
## Reason
When a user really really wants to specify the `id` field in their queries, this is possible with the `allow_user_specified_id` parameter. 
However, when generating the code, this is not possible to specify, which causes the CLI codegen to fail:
```
gel.errors.QueryError: cannot assign to property 'id'
   ┌─ query:5:7
   │ 
 5 │         id := <uuid>$entity_id
   │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ consider enabling the "allow_user_specified_id" configuration parameter to allow setting custom object
```
Since edge cases exist where it might be beneficial to specify `id` in queries I added a `--allow-user-specified-id` parameter to the CLI which sets this option on the client used for the query codegen.

## Testing
To test this I quickly replaced my `gel` dependency with my fork, it works as expected. If required please let me know how you would like an additional unit test for this to look like.

### DIFF
--- a/gel/codegen/cli.py
+++ b/gel/codegen/cli.py
@@ -77,7 +77,6 @@ parser.add_argument(
 parser.add_argument(
     "--allow-user-specified-id",
     action=argparse.BooleanOptionalAction,
-    default=argparse.SUPPRESS,  # override the builtin help for default
     help="Allow user specified ids in .edgeql files (default is to disallow).",
 )
 
--- a/gel/codegen/cli.py
+++ b/gel/codegen/cli.py
@@ -28,8 +28,7 @@ class ColoredArgumentParser(argparse.ArgumentParser):
         c = generator.C
         self.exit(
             2,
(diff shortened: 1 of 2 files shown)


gel/codegen/cli.py
```diff
@@ -75,6 +74,12 @@ def error(self, message):
     help="Add a mixin to generated dataclasses "
     "to skip Pydantic validation (default is to add the mixin).",
 )
+parser.add_argument(
+    "--allow-user-specified-id",
+    action=argparse.BooleanOptionalAction,
+    default=argparse.SUPPRESS,  # override the builtin help for default
```
fantix, line 80:
```suggestion
```

We don't want to suppress the default `False` here as it is a forward binary option.

fantix (changes requested):

PR #640, opened 2025-06-01
FastAPI integration with Gel Auth

fantix:
- [x] `gel.fastapi.gelify()` for opinionated Gel client integration with FastAPI
- [x] Support globals
- [x] `gel.fastapi.gelify().auth` for opinionated Gel Auth integration with FastAPI
- [x] Add default error logging
- [x] Add default redirect
- [x] Add `on_new_identity`
- [x] Add built-in UI
- [x] Merge `authify`
- [x] Add safeguards
- [x] Add support of adaptive sync/async callbacks
- [x] Add support for DI
- [x] Add async/non-async clients
- [x] Move internals
- [x] Cache auth_token dependency

### Basic client integration:

```python
import fastapi, gel.fastapi

app = fastapi.FastAPI()
gel.fastapi.gelify(app)

@app.get("/")
async def index(client: gel.fastapi.Client):
    return str(await client.query("select 40 + 2"))
```

In non-async functions, you must use `BlockingIOClient`:

```python
@app.get("/")
def index(client: gel.fastapi.BlockingIOClient):
    return str(client.query("select 40 + 2"))
```

Or use the underlying raw client directly/globally:

```python
g = gel.fastapi.gelify(app)

@app.get("/")
async def index():
    return str(await g.client.query("select 40 + 2"))

@app.get("/")
def index():
    return str(g.blocking_io_client.query("select 40 + 2"))
```

Note, `g.client` is guaranteed to come with no auto-globals, while the injected dependency`client: gel.fastapi.Client` is context-aware, see the globals section below.

### Customization

This example includes lifespan customization and client options:

```python
import fastapi, gel.fastapi

g = gel.fastapi.gelify(
    app,
    shutdown_timeout=5,
    state_name="geldb",
).with_client_options(
    dsn="gel://1.2.3.4:5656/mydb",
    wait_until_available=3,
    timeout=6,
)

@app.get("/")
async def index(request: fastapi.Request):
    return str(await request.state.geldb.query("select 40 + 2"))
```

Then `request.state.geldb` becomes the context-aware version of `g.client`. Note, `with_client_options()` must be used before accessing `g.client` or FastAPI application startup, or it will raise a `ValueError`.

### Globals with raw FastAPI security:

```python
import typing, fastapi.security, gel.fastapi

app = fastapi.FastAPI()
g = gel.fastapi.gelify(app)
security = fastapi.security.HTTPBearer()

@g.with_global("user::token")
def user_token(
    credentials: typing.Annotated[
        fastapi.security.HTTPAuthorizationCredentials,
        fastapi.Depends(security),
    ],
) -> str:
    return credentials.credentials

@app.get("/secure", dependencies=[user_token])
async def secure(client: gel.fastapi.Client) -> str:
    return str(await client.query("select global user::token"))
```

You can also put the dependency on a `fastapi.Router` to protect a group of endpoints:

```python
user_router = fastapi.Router(dependencies=[user_token])
```

### Gel Auth:

```python
import random, typing, uuid
import fastapi.responses, gel.fastapi, gel.auth

"""
using extension auth;

module default {
    type User {
        name: str;
            constraint exclusive;
        };
    }
}
"""

app = fastapi.FastAPI()
g = gel.fastapi.gelify(app)


async def sign_in_page(
    client: gel.fastapi.Client,
    incomplete: typing.Optional[str] = None,
) -> str:
        """
        select (
            select User
        ).name;
        """
    )
    return incomplete or "hello, %s" % (name or "stranger")


@app.get("/error")
def error_page(error: str) -> str:
    return error


@app.get("/reset-password")
def reset_password_page() -> str:
    return "Reset your password here"


@g.auth.on_new_identity
async def on_new_identity(
    result: tuple[uuid.UUID, typing.Optional[gel.auth.TokenData]],
    client: gel.fastapi.Client,
) -> None:
    identity_id, token_data = result
        """
        insert User {
            name := <str>$name,
        }
        """,
        name=random.randbytes(8).hex(),
        identity_id=identity_id,
    )
```

By default, the minimum steps to use the Gel Auth built-in email+password integration are:

* Use the Gel `auth` extension in the schema
* Add the email+password provider
* Define a `/` route as the default redirect_to
* Define a route named `error_page` to show auth errors

The "Gel Auth" endpoints are added based on the result of:

```edgeql
select assert_single(
    cfg::Config.extensions[is ext::auth::AuthConfig]
) {
    providers: { id, name },
    ui,
}
```

The `on_...` hooks are handled in the same way FastAPI handles routes, so they support blocking I/O out of the box:

```python
@g.auth.on_new_identity
def on_new_identity(
    result: tuple[uuid.UUID, typing.Optional[gel.auth.TokenData]],
    client: gel.fastapi.BlockingIOClient,
) -> None:
    identity_id, token_data = result
    client.query_required_single(
        """
        insert User {
            name := <str>$name,
        }
        """,
        name=random.randbytes(8).hex(),
        identity_id=identity_id,
    )
```

Note the `gel.fastapi.BlockingIOClient` part too.

You can also use some of the familiar FastAPI perks:

```python
@g.auth.on_new_identity(
    response_class=responses.RedirectResponse,
    status_code=http.HTTPStatus.SEE_OTHER,
):
def on_new_identity(
    result: ...,
    someone: Annotated[str, fastapi.Depends(my_dep)],
):
    return f"/welcome?who={someone}"
```

### Extension Customization

This integration has many customizable configs. Configs and hooks can be set only before the FastAPI application starts to run. For example:

```python
def on_new_identity(...): ...

gel.fastapi.gelify(app).with_auth(
    auth_cookie_name="token",
    verifier_cookie_name="pkce",
    secure_cookie=False,
    on_new_identity=on_new_identity,
)
```

Or identically with full typing support:

```python
g = gel.fastapi.gelify(app)
(
    g.auth
    .auth_cookie_name("token")
    .verifier_cookie_name("pkce")
    .secure_cookie(False)
)

@g.auth.on_new_identity
def on_new_identity(...): ...
```

### Swagger UI example

image


1st1:
>  gel.fastapi.gelify() for opinionated Gel client integration with FastAPI

This is pretty cool, but I've seen people using `def` and `async def` handlers in FastAPI projects -- kinda terrifying -- but even the official tutorial has those. I'm really not sure what to do about that. :/

msullivan:
Agree about merging `gelify` and `authify`.

Looking at some tutorials, both global variable based patterns and Depends based patterns seem to be idiomatic with fastapi.
I *think* it would be best if we don't force using the global, though.

It should definitely be possible to have a `Client` that uses `Depends` to automatically find a client.
I *think* something like:
```
async def get_client(request: Request) -> gel.AsyncIOClient:
    return request.app.state.gel_client

Client = typing.Annotated[gel.AsyncIOClient, fastapi.Depends(get_client)]
```

We could also make `get_client` lazily create a client even if we haven't called `gelify`, if we want. (@1st1, is that what you were suggesting?)

I'm not sure if we could make it use the "lifespan" mechanism if we do that, and I think it probably wouldn't be reasonable (or possible?) to automagically enable auth stuff, since that needs to serve routes.

1st1:
> We could also make get_client lazily create a client even if we haven't called gelify, if we want. (@1st1, is that what you were suggesting?)

I like it! @fantix thoughts? How far can we push this all?

fantix:
> > We could also make get_client lazily create a client even if we haven't called gelify, if we want. (@1st1, is that what you were suggesting?)
> 
> I like it! @fantix thoughts? How far can we push this all?

This would probably fall into the Zero Call idea where we may need to monkey patch the `FastAPI()` constructor when importing `gel` itself (or `gel.fastapi`) in order to install all routes at the starting of the Web server, or it may remain as just a one-off `Client`, which is also fine, probably:

```python
async def one_off_client():
    client = create_async_client()
    try:
        yield client
    finally:
        await client.aclose()

AsyncIOClient = Annotated[gel.AsyncIOClient, fastapi.Depends(one_off_client)]
```

msullivan:
One off client is bad, since it would require a client that maintains singletons under the hood to perform ok.

My theory was was that we'd still stash the client in `request.app.state` and then maybe use the (deprecated but probably too useful to actually ever get removed?) `on_event('shutdown')` to clean it up.

I think that trying to make auth work like this is probably too much magic to be palatable, and we shouldn't do it.

Honestly I kind of think that making get_client lazily make a client at all is mostly a gimmick too, but we might be looking for gimmicks.

fantix:
> `on_event('shutdown')` to clean it up

`on_event('shutdown')` won't work; lifespan is pre-calculated and its `__aexit__` is cached by an app-lifetime-global `async with`. I've already tried many hacky ways, it seems impossible to hook in clean-ups at runtime properly into the ASGI lifetime, and monkey-patching `FastAPI` at as early as import-time (the Zero Call idea) seems to be the only way.

Another possibility to use Gel without `gelify()` is a [global dependency](https://fastapi.tiangolo.com/tutorial/dependencies/global-dependencies/), it's like the one-off client example above, but applied to the root router for all routes:

```python
import fastapi, gel.fastapi

app = fastapi.FastAPI(dependencies=[gel.fastapi.one_off_client])

@app.get("/")
async def index(client: gel.fastapi.AsyncIOClient):
    await client.query("select 42")
```

I wouldn't suggest this, because it's not simpler than just `gel.fastapi.gelify(app)`.

**My Conclusion:** unless we go for the monkey-patch-based Zero Call idea, the user will have to use `gelify()`. But at the very least, I added a built-in dependency:

```python
import fastapi, gel.fastapi

app = fastapi.FastAPI()
gel.fastapi.gelify(app)


@app.get("/")
async def index(client: gel.fastapi.Client):
    return str(await client.query("select 42"))
```

1st1:
Wow, massive and impressive work.

* Re `gel.fastapi.gelify(app, wait_until_available=3, timeout=6)
` -- I'm wondering if we're painting ourselves in the corner here. We might want `gelify` to have own arguments eventually. How about `gel.fastapi.gelify(app).with_client_options(wait_until_available=3, timeout=6)` or something like that?

* Re having both `gel.fastapi.gelify(app).with_auth(` and `gel.fastapi.gelify(app).auth` -- can we have just one?

----

So, can this happen:

```python

g = (gel.fastapi.gelify(app)
     .wait_until_available=3, timeout=6).
     .with_auth(...)
     )

@g.auth.on_new_identity
def on_new_identity(...): ...
```

? *I'd love this kind of simplicity.*

Right now I'm feeling we have too many magical knobs and I wish we had just one builder pattern.

1st1:
@fantix huge thanks! <3

PR #667, opened 2025-06-09
A better save() implementation

1st1:


1st1:
CleanShot 2025-06-09 at 08 00 24@2x
Thanks @elprans !

PR #680, opened 2025-06-10
models: Add support for putting literal '*' in `select`

elprans:
`Foo.select("*")` is sugar for `Foo.select(*Foo)`

Additionally, polymorphic splats are supported.


gel/_internal/_qbmodel/_abstract/_expressions.py
```diff
@@ -90,23 +90,22 @@ def _const_to_expr(
 
 
 def select(
-    cls: type[GelType],
+    cls: type[GelObjectType],
     /,
-    *elements: _qb.PathAlias,
+    *elements: _qb.ShapeElement | Literal["*"],
```
1st1, line 95: do we want to add `**` now? i'm pretty sure people will be asking for it

1st1 (approved):
approved but please fix object.pyx.

PR #682, opened 2025-06-11
models: Exclude computeds from `*` splats

elprans:


elprans:
Decided not to do this.

PR #687, opened 2025-06-12
Assorted fixes

elprans:
- **codegen/models: Allow specifying output directory**
- **fastapi: Patch `gel watch` into `fastapi dev` when gelify is used**
- **fastapi.auth: Check for presence of `auth` extension before doing anything**

### DIFF
--- a/gel/_internal/_integration/_fastapi/_auth/__init__.py
+++ b/gel/_internal/_integration/_fastapi/_auth/__init__.py
@@ -8,6 +8,7 @@ from typing_extensions import Self
 
 import contextlib
 import datetime
+import logging
 
 import fastapi
 import jwt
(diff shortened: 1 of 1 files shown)


gel/_internal/_integration/_fastapi/_auth/__init__.py
```diff
@@ -239,6 +243,22 @@ async def on_startup(self, app: fastapi.FastAPI) -> None:
...
+                """
+                select assert_single(
+                    (select schema::Extension filter .name = "auth")
+                )
+                """
+            )
+
+            if not ext:
+                _logger.warning(
+                    "auth extension not installed, add `use extension auth;` "
```
fantix, line 256: typo:

```suggestion
                    "auth extension not installed, add `using extension auth;` "
```

Issue #691, opened 2025-06-12
Filter by UUID doesn't work


anbuzin:
```python
post = await db.get(default.Post.filter(id=uuid.UUID("68ed4c68-478e-11f0-8af4-aff8c8f90955")))
# AttributeError: 'UUID' object has no attribute 'replace'
```

Doesn't work with `std.uuid` either.

Issue #690, opened 2025-06-12
Filter by boolean seems broken


anbuzin:
```python
posts = default.Post.filter(published=True)
# ValueError: get_literal_for_value: 1 is not an instance of <class 'bool'>
```

I tried the lambda options too, but they didn't work either:

```python
posts = default.Post.filter(lambda p: p.published is True)
# TypeError: bool cannot be converted to an Expr
```

PR #696, opened 2025-06-13
models: Fix raw value scalar casting

elprans:
Implicit and explicit casting of base Python scalars to reflected
scalars should now work.

Fixes: #690
Fixes: #691


gel/_internal/_qbmodel/_abstract/_primitive.py
```diff
@@ -403,26 +430,206 @@ def get_py_type_scalar_match_rank(
...
+    def __edgeql_literal__(self) -> _qb.Literal | _qb.CastOp:
+        return get_literal_for_scalar(type(self), self)
+
+
+UUIDFieldsTuple = TypeAliasType(
+    "UUIDFieldsTuple", tuple[int, int, int, int, int, int]
+)
+
+
+class UUIDImpl(uuid.UUID):
```
1st1, line 476: why do we need these wrappers?

gel/_internal/_qbmodel/_abstract/_primitive.py
```diff
@@ -349,6 +351,31 @@ def _get_py_type_for_abstract_scalar(typename: str) -> set[tuple[str, str]]:
     return union
 
 
+MODEL_SUBSTRATE_MODULE: Final = "__gel_substrate__"
```
1st1, line 354: would love to have a comment here explaining what substrate means and  what the attribute is supposed to do.

1st1 (approved):

gel/_internal/_qbmodel/_abstract/_primitive.py
elprans: To allow downcasting, i.e `std.uuid(uuid.UUID(...))`, because standard `uuid.UUID` class does not support this.

gel/_internal/_qbmodel/_abstract/_primitive.py
elprans: added

Issue #688, opened 2025-06-12
Model generation breaks if you run it more than once


anbuzin:
When I run `gel-generate-python models` twice in a row it breaks imports:

```
Traceback (most recent call last):
  File "/Users/andrey/local/projects/python-api-project/main.py", line 3, in <module>
    from models import default, std
  File "/Users/andrey/local/projects/python-api-project/models/__init__.py", line 9, in <module>
    from .default import Comment, Post, Text, User
ModuleNotFoundError: No module named 'models.default'
```

PR #702, opened 2025-06-16
models: Do not rsync the models directory when no generation was made

elprans:
If no actual generation wasn't made we shouldn't attempt to dirsync the
models directory, that's pointless and will also erase user models.

Fixes: #688


1st1:
> that's pointless and will also erase user models.

What does this mean? Are we encouraging users to put their code in `/models`? Should they? Why would it erase anything?

1st1 (approved):

Issue #670, opened 2025-06-09
Expose operators in ORM


vpetrovykh:
The ORM should have a way to use `like`, `ilike`, `and`, `or`, `not` in expressions (select computeds, updates, filters, order by, etc.)

It probably makes sense to just add these as functions to the reflected `std`. So we can have:
```python
q0 = default.User.filter(lambda u: std.ilike(u.name, 'a%')
q1 = default.User.filter(lambda u: std.or_(u.name == 'Alice', u.name == 'Billie')
```
`or_`, `and_`, `not_` will need the `_` because they are reserved keywords in Python and that's the common Pythonic practice.

vpetrovykh:
There are also `in`, `not in` so that it's possible to do things like:
```
q2 = default.User.filter(lambda u: std.in(u.name in {'Alice', 'Billie'}))
q3 = default.User.filter(lambda u: std.not_in(u.name, u.friends.name))
q4 = default.User.filter(lambda u: std.in('Alice', u.friends.name))
```

vpetrovykh:
Coalescing operators (`??`, `?=`, `?!=`) and `if ... then ... else ...` will need their own reflections.
As for naming we already have an annotation `identifier` associated with the coalescing operators, so they can be: `std.coalesce`, `std.coal_eq`, `std.coal_neq`. Similarly, given the `identifier` annotation for the `IF` it can be reflected as `std.if_else`.

PR #704, opened 2025-06-16
models: Reflect non-magic operators as well

elprans:
Add reflection of EdgeQL operators that have no corresponding Python
dunder magic.  These are reflected as overloaded functions in std with
names specified by the `std::identifier` annotation or the operator name
if its an identifier itself.  Operators that are clashing with Python
keywords (e.g `or`) are qualified with trailing underscore (`or_()`).

Fixes: #670


1st1 (approved):

1st1:
(only looked at this briefly)

Issue #706, opened 2025-06-17
Can't import ext::ai reflection


anbuzin:
```python
from models.ext import ai
```

```
TypeError: cannot specify both default and default_factory
```

elprans:
Should have been fixed in #704

anbuzin:
Tried it with latest master, unfortunately the error is still there.

1st1:
I can't repro this specific error @anbuzin 

Are you sure you got

> TypeError: cannot specify both default and default_factory

with specifically this schema? Maybe `uv` is caching the package for you and you didn't get a fresh build of master?

1st1:
There are other errors though, so I'll try to fix them

anbuzin:
Ok so this exact schema in a brand new project on 6.8 and

```
 + gel==3.1.0 (from git+https://github.com/geldata/gel-python.git@60800748801dd5ebac89a185785a6f99578a99e5)
```

Gives this error:

```
(orm-repro) andrey@mac orm-repro % python3 main.py
Traceback (most recent call last):
  File "/Users/andrey/local/projects/extension_ai_polished/orm-repro/main.py", line 1, in <module>
    from models.ext import ai
  File "/Users/andrey/local/projects/extension_ai_polished/orm-repro/models/ext/ai.py", line 10, in <module>
    from ..__variants__.ext import ai as base
  File "/Users/andrey/local/projects/extension_ai_polished/orm-repro/models/__variants__/ext/ai.py", line 984, in <module>
    class Config(
    ...<164 lines>...
            pass
  File "/Users/andrey/local/projects/extension_ai_polished/orm-repro/models/__variants__/ext/ai.py", line 1059, in Config
    class __variants__(cfg.ExtensionConfig.__variants__):
    ...<85 lines>...
        Any = TypeVar("Any", bound="Config | Base | Required | Partial")
  File "/Users/andrey/local/projects/extension_ai_polished/orm-repro/models/__variants__/ext/ai.py", line 1140, in __variants__
    class Partial(PartialBase, cfg.ExtensionConfig.__variants__.Partial):  # type: ignore [misc, unused-ignore]
        indexer_naptime: OptionalProperty[___std_1__.duration, timedelta]
        providers: MultiLink[___ext_ai__.ProviderConfig | ___ext_ai__.ProviderConfig.__variants__.Partial]
[...]

PR #712, opened 2025-06-18
models: Handle variant unions in link annotations

elprans:
Fixes: #706

### DIFF
--- a/gel/_internal/_qbmodel/_abstract/_base.py
+++ b/gel/_internal/_qbmodel/_abstract/_base.py
@@ -105,7 +105,15 @@ class GelObjectType(
     _qb.GelObjectTypeMetadata,
     metaclass=GelObjectTypeMeta,
 ):
-    pass
+    __gel_variant__: ClassVar[str | None] = None
+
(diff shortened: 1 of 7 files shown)


gel/_internal/_qbmodel/_abstract/_base.py
```diff
@@ -105,7 +105,15 @@ class GelObjectType(
     _qb.GelObjectTypeMetadata,
     metaclass=GelObjectTypeMeta,
 ):
-    pass
+    __gel_variant__: ClassVar[str | None] = None
```
1st1, line 108: I don't understand the need of `__gel_variant__`. Is it just a marker on a type?

elprans: Though actually, it would be quite annoying to have it as an enum, because variants are inherently target-specific (e.g Pydantic) and the field is defined in the "abstract" query-builder model.  I'd leave it as a string for now.

1st1 (approved):
Please merge so that @anbuzin can play with the ai side of things. Comments would be appreciated.

PR #714, opened 2025-06-18
batch: consider the error of wait() on __exit__

fantix:
* Don't automatically `wait()` on `__exit__` if there was an error already
* The automatic `wait()` can now trigger retry too
* Also, disable `session_idle_transaction_timeout` when running batch

### DIFF
--- a/gel/asyncio_client.py
+++ b/gel/asyncio_client.py
@@ -369,7 +369,16 @@ class AsyncIOBatchIteration(transaction.BaseTransaction):
         return self
 
     async def __aexit__(self, extype, ex, tb):
-        await self.wait()
+        if extype is None:
+            try:
+                await self.wait()
+            except Exception as ex:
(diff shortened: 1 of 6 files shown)


gel/asyncio_client.py
```diff
@@ -515,7 +534,11 @@ def transaction(self) -> AsyncIORetry:
         return AsyncIORetry(self)
 
     def _batch(self) -> AsyncIOBatch:
-        return AsyncIOBatch(self)
+        return AsyncIOBatch(
+            self.with_config(
+                session_idle_transaction_timeout=datetime.timedelta()
```
1st1, line 539:
```suggestion
# we only need to specify transaction idle timeout,
# session idle timeouts can't interrupt transactions                session_idle_transaction_timeout=datetime.timedelta()
```

1st1 (approved):

PR #726, opened 2025-06-23
In gel.ai print error messages from the server

anbuzin:


gel/ai/core.py
```diff
@@ -147,7 +147,13 @@ def query_rag(
                 stream=False,
             ).to_httpx_request()
         )
-        resp.raise_for_status()
+        try:
+            resp.raise_for_status()
+        except httpx.HTTPStatusError as e:
+            error = e.response.json()
+            print(f"{error['type']}: {error['message']}")
+            raise
```
fantix, line 155:
```suggestion
        if resp.is_error:
            logger.error("HTTP error: %(type)s: %(message)s", resp.json())
            resp.raise_for_status()
```

PR #730, opened 2025-06-23
refactor: smart gelify

fantix:
This PR does more automatic checks and behaves more reasonably:

* If `g.without_auth()` is called, the auth endpoints are guaranteed not to be installed in any cases
* Or else, if `g.with_auth()` is called or `g.auth` is accessed, the auth extension is considered *manually* enabled
  * At application startup, we will introspect to see if the `auth` extension is actually enabled in the Gel schema, and raise an error if it's not:
    > Extension "auth" is not installed, add `use extension auth;` to your Gel schema to enable it.
* In the end, if the user only called `gelify()` and nothing else about the auth extension,  we go 100% automatic:
  * If the `auth` extension is not enabled in the Gel schema, we silently skip the auth extension at all
  * Or else, we check for the necessary Python dependencies (`httpx` and `pyjwt` in this case). If any of them is not installed, we skip the auth extension with a *warning*:
    > Required dependencies ["xx", "yy"] are not installed. Please install them to use the "GelAuth" extension, or disable the extension to avoid this warning.
* The `python-multipart` Python dependency is now made optional and automatic, the default auth routes will support both `application/json` and `application/x-www-form-urlencoded` request content type if `python-multipart` is installed; or else, only `application/json` is allowed.
* (no change in this PR) The auth extension currently has 2 installables: `email_password` and `builtin_ui`. They work pretty much the same way as extensions. In automatic mode, we look at the schema and enable what's enabled.

### DIFF
--- a/gel/_internal/_integration/_fastapi/_auth/__init__.py
+++ b/gel/_internal/_integration/_fastapi/_auth/__init__.py
@@ -8,7 +8,6 @@ from typing_extensions import Self
 
 import contextlib
 import datetime
-import logging
 
 import fastapi
 import jwt
@@ -27,9 +26,6 @@ if TYPE_CHECKING:
     from gel import auth as core
     from .email_password import EmailPassword
     from .builtin_ui import BuiltinUI
-
-
-_logger = logging.getLogger("gel.fastapi.auth")
 
 
 class Installable:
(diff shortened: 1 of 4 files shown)


scotttrinh (approved):
👏

PR #732, opened 2025-06-24
fastapi-auth: fix built-in UI callbacks

fantix:
Changes:

* Moved `redirect_to`/`redirect_to_page_name` to `g.auth`
* Add default auth callback handlers
* on_xxx_complete will set the auth token in cookie
* on_xxx_complete hooks will be called with token global

Example application:

```python
import random, typing, uuid
import fastapi.responses, gel.fastapi, gel.auth

"""
using extension auth;

module default {
    type User {
        name: str;
            constraint exclusive;
        };
    }
}
"""

app = fastapi.FastAPI()
g = gel.fastapi.gelify(app)


async def root(
    client: gel.fastapi.Client,
    request: fastapi.Request,
):
        """
        select (
            select User
        ).name;
        """
    )
    if name is None:
        return fastapi.responses.RedirectResponse(
            request.url_for(g.auth.builtin_ui.sign_in_name.value)
        )
    else:
        return "hello, %s" % name


@g.auth.on_new_identity
async def on_new_identity(
    result: tuple[uuid.UUID, typing.Optional[gel.auth.TokenData]],
    client: gel.fastapi.Client,
) -> None:
    identity_id, token_data = result
        """
        insert User {
            name := <str>$name,
        }
        """,
        name=random.randbytes(8).hex(),
        identity_id=identity_id,
    )
```

Requires Gel configuration: `ext::auth::UIConfig.redirect_to` and `redirect_to_on_signup` must be set with:
* Both absolute URL
* Identical URL scheme and net location
* Identical path, where the `redirect_to_on_signup` must have query `?isSignUp=true`; or different paths


scotttrinh (approved):
Confirmed that this works via manual testing.

elprans (approved):

PR #733, opened 2025-06-25
fastapi-auth: allow more customization

fantix:
Added:
* Ability to derive `SignUpBody`
* Ability to turn off individual default routes included

Changed:
* Hook arguments are injected now, no more tuples

Full example to allow sign-up with name:

```python
import typing, fastapi, gel.fastapi.auth.email_password as ep, gel.auth

"""
using extension auth;

module default {
    type User {
        name: str;
            constraint exclusive;
        };
    }
}
"""

app = fastapi.FastAPI()
g = gel.fastapi.gelify(app)


@g.auth.email_password.sign_up_body
class SignUpBody(ep.SignUpBody):
    name: str


async def root(client: gel.fastapi.Client) -> str:
        """
        select (
            select User
        ).name;
        """
    )
    return "hello, %s" % (name or "guest")


@g.auth.email_password.on_sign_up_complete
async def sign_up_complete(
    body: SignUpBody,
    result: gel.auth.email_password.SignUpCompleteResponse,
    client: gel.fastapi.Client,
    request: fastapi.Request,
) -> str:
        """
        insert User {
            name := <str>$name,
        }
        """,
        name=body.name,
        identity_id=result.identity_id,
    )
    return str(request.url_for("root"))
```

### DIFF
--- a/gel/_internal/_integration/_fastapi/_utils.py
+++ b/gel/_internal/_integration/_fastapi/_utils.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 from typing import (
     Annotated,
     Any,
+    cast,
     Generic,
     get_args,
     get_origin,
@@ -46,6 +47,7 @@ class ConfigSubject(Protocol):
     installed: bool
 
 
+C = TypeVar("C")
 S = TypeVar("S", bound=ConfigSubject)
 T = TypeVar("T")
 Ts = TypeVarTuple("Ts")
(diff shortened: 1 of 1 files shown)


fantix:
> We should make it clear somehow what order the `on_new_identity` and individual on_sign_up methods work. A somewhat generic API server might want to define `on_new_identity` to create their `User` object, but define how to get `full_name` and `email` based on the authentication strategy, but we'd need to ensure that `on_new_identity` is always run first so there is a user to associate with your user data.

Right, `on_new_identity` is guaranteed to run before other hooks, if and only if a new `Identity` is created. However, the hooks will run their own transactions, with which I feel unsafe to separate the database logic. It's better to use separate `on_sign_up` hooks with a shared function to create the identity in the same transaction; `on_new_identity` on the other hand can be used as an alternative when all successful sign-ups should return the same result.

(Even though, it's still possible and up to the user to separate logic in different transactions, afterall the server-side identity creation is already a separate transaction anyways. It's just tricky to put this pattern together correctly, I'm sure a large percentage of users will miss handling one or two edge cases, leaving a broken state behind, like an email would never be able to be used to sign in/up without admin's help. We should have more code/docs to make this less wrong.)

PR #741, opened 2025-06-27
Add test for updating a model with model_copy

scotttrinh:
This seems to be a common way of applying a sparse set of updates to an existing object. See https://fastapi.tiangolo.com/tutorial/body-updates/#partial-updates-with-patch


1st1:
Part of https://github.com/geldata/gel-python/pull/738 now

PR #742, opened 2025-06-27
Exclude unset .id from model_dump / model_dump_json

1st1:


gel/_internal/_utils.py
```diff
@@ -23,3 +35,18 @@ def type_repr(t: Any) -> str:
             return f"{t.__module__}.{t.__qualname__}"
     else:
         return repr(t)
+
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+def inherit_signature(
```
elprans, line 44: Nice hack

gel/_internal/_qbmodel/_pydantic/_models.py
```diff
@@ -383,6 +389,24 @@ def __delattr__(self, name: str) -> None:
...
+        try:
+            exclude = kwargs["exclude"]
+        except KeyError:
+            exclude = kwargs["exclude"] = set()
+
+        if isinstance(exclude, set):
+            exclude.add("id")
+        else:
+            assert isinstance(exclude, dict)
+            exclude["id"] = True
```
elprans, line 405: Slightly more optimal, since "exclude" will not be there most of the time:

```suggestion
        try:
            exclude = kwargs["exclude"]
        except KeyError:
            kwargs["exclude"] = {"id"}
        else:
            if isinstance(exclude, set):
                exclude.add("id")
            else:
                assert isinstance(exclude, dict)
                exclude["id"] = True
```

elprans (approved):

gel/_internal/_qbmodel/_pydantic/_models.py
1st1: ++

PR #746, opened 2025-07-01
models: Improve handling of user-defined model variants

elprans:
User-defined model variants are now handled better.  Any ad-hoc pointer
definitions are automatically marked as computed and non-global class
definitions are properly supported by patching
`__pydantic_parent_namespace__`.


1st1 (approved):
💀💀💀

PR #751, opened 2025-07-03
Allow passing 'id' to model __init__

1st1:


gel/_internal/_codegen/_models/_pydantic.py
```diff
@@ -3480,23 +3539,55 @@ def _write_base_object_type_body(
...
-                    "Call db.save() on the returned object to persist it "
-                    "in the database."
-                )
-                self.write('"""')
+            render_id_variant = False
+
+            if (
+                objtype.name != "std::FreeObject"
+                and not objtype.name.startswith("schema::")
+                # and not objtype.name.startswith("std::")
```
elprans, line 3594: Any reason why this is commented out?

tests/test_model_generator.py
```diff
@@ -3395,7 +3395,7 @@ def test_modelgen_reflection_1(self):
             "tuple<a:std::str, b:tuple<c:std::int64, d:std::str>>",
         )
 
-    @tb.typecheck
+    # @tb.typecheck
```
elprans, line 3398: We should probably have a `typecheck_xfail` decorator

1st1: ok

PR #753, opened 2025-07-04
models: Fix import cycles

elprans:
`late_runtime` imports currently used do not guarantee absence of import
cycles, so introduce a new `ImportTime.typecheck_runtime` which is a
hybrid of `if TYPE_CHECKING` import and a `DeferredImport` mechanism
that defers module import until first attribute access.  The intent of
`typecheck_runtime` is to allow ForwardRefs work both at type-checking
tima and at type-eval time, so `DeferredImport` converts import and
attribute errors triggered by partially initialized modules into
`NameError`s understood by type eval.

### DIFF
--- a/gel/_internal/_deferred_import.py
+++ b/gel/_internal/_deferred_import.py
@@ -118,7 +118,7 @@ class DeferredImport:
 
             >>> # Equivalent to: from .parent import submodule
             >>> submodule = DeferredImport('parent', attr='submodule',
-                                          package=__name__)
+                                           package=__package__)
 
         Note:
             The actual import is deferred until the first attribute access on
--- a/gel/_internal/_deferred_import.py
+++ b/gel/_internal/_deferred_import.py
@@ -1,0 +1,197 @@
+# SPDX-PackageName: gel-python
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright Gel Data Inc. and the contributors.
+
+from __future__ import annotations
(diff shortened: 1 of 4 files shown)


gel/_internal/_deferred_import.py
```diff
@@ -0,0 +1,197 @@
...
+        if (caller_module := _get_caller_module()) is not None:
+            self.__lm_importing_module__ = weakref.ref(caller_module)
+        else:
+            self.__lm_importing_module__ = None
+
+    def __load(self) -> types.ModuleType:
+        if (mod_ref := self.__lm_module__) is not None and (
+            mod := mod_ref()
+        ) is not None:
+            return mod
```
1st1, line 156: Should you replace itself with the resolved object once the real import is done?

1st1:
Oh we need a regression test in some form for this!

gel/_internal/_deferred_import.py
elprans: This is literally done a few lines down.  If we couldn't write to caller's globals when the actual import happened then we don't try again.

elprans:
> Oh we need a regression test in some form for this!

The import cycle started triggering for me on `test_modelgen_empty_ai_schema_1` (with changes in an upcoming PR).  The stdlib is already full of cycles, it's a small miracle things were working as well so far.

gel/_internal/_deferred_import.py
1st1: ok I've missed it, cool

Issue #651, opened 2025-06-02
Make sure `gel-generate models` produces stable code


1st1:
Codegen should alphabetically sort types, user-defined functions, and imports.

Properties and links should be in the order they are defined in the schema within the type? (IIRC it should be stable, but doublecheck, and if not - alphabetically sort link/properties too)

PR #770, opened 2025-07-09
models: Tweaks to make generator output deterministic

elprans:
Add a test to catch regressions

Fixes: #651

### DIFF
--- a/gel/_internal/_reflection/_callables.py
+++ b/gel/_internal/_reflection/_callables.py
@@ -72,6 +72,10 @@ class CallableParam:
             return self.index
 
     @functools.cached_property
+    def sort_key(self) -> str:
+        return str(self.key)
+
+    @functools.cached_property
     def is_variadic(self) -> bool:
         return self.kind is CallableParamKind.Variadic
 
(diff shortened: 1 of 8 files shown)


tests/test_model_generator.py
```diff
@@ -4448,3 +4453,60 @@ def test_modelgen_escape_02(self):
...
+                        "models",
+                        "--output=models",
+                    ],
+                    cwd=td,
+                    env=env,
+                    check=True,
+                )
+
+                if i > 0:
+                    diff = _dirdiff.unified_dir_diff(
```
1st1, line 4503: nice

gel/_internal/_reflection/_callables.py
```diff
@@ -71,6 +71,10 @@ def key(self) -> CallableParamKey:
         else:
             return self.index
 
+    @functools.cached_property
+    def sort_key(self) -> str:
+        return str(self.key)
```
1st1, line 76: when in CI you can also mix hashseed right in

1st1 (approved):

gel/_internal/_reflection/_callables.py
elprans: What do you mean?  This sort key should be stable regardless of seed.

PR #775, opened 2025-07-10
Work on support JSON schemas

1st1:
Fix JSON schemas / serialization / validation / and dump exclude

This lacks tests they will come later.


gel/_internal/_dlist.py
```diff
@@ -37,6 +36,9 @@
 from gel._internal import _typing_parametric as parametric
 from gel._internal._polyfills._strenum import StrEnum
 
+if TYPE_CHECKING:
+    from gel._internal._qbmodel._pydantic import GelModel
```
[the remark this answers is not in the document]
elprans: Also, since we're tying `_dlist.py` to the model like that it probably makes sense to just move it to `_qbmodel/_abstract/` now.

gel/_internal/_dlist.py
```diff
@@ -516,6 +518,15 @@ def __init__(
...
         )
 
+    def _check_value(self, value: Any) -> _MT_co:
+        cls = type(self)
+
+        t = cls.type
+        if isinstance(value, cls.type):
+            return value
+
+        return t.model_validate(value)
```
elprans, line 528: Here you can defer to some dunder in the abstract model that, in turn, calls `model_validate` in its `_pydantic.GelModel` overload.

1st1: I don't understand what you suggest

1st1: ok done

gel/_internal/_dlist.py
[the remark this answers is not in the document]
1st1: done

PR #777, opened 2025-07-11
qb: Add support for globals

elprans:
Globals are reflected as values of their type and can be referenced in
query builder expressions simply by name:

    client.query(models.MyGlobal)


tests/test_model_generator.py
```diff
@@ -3549,6 +3549,47 @@ def test_modelgen_linkprops_3(self):
...
+        self.assertTrue(
+            sess_client.get(default.current_game_session_num < 1000)
+        )
+
+        # That object globals work also
+        fetched_sess = sess_client.get(default.CurrentGameSession)
+        self.assertEqual(fetched_sess, sess)
+
+        fetched_sess = sess_client.get(
+            default.CurrentGameSession.filter(
```
1st1, line 3587: Can you add a test that selects a shape out of a global

1st1 (approved):

Issue #779, opened 2025-07-11
fix linkprops


1st1:
1. kill `__lprops__`.
  * `model_instance.__linkprops__` should stay as is
  * `ModelType.__linkprops__` should replace `ModelType.__lprops__`

2. Make sure we always fetch *all* link properties for **link and link computeds**. No matter the query, be it explicit select shape, no `.select()` at all, or any other possible way. If we don't, `save()` will remove data unknowingly to the user.

3. @vpetrovykh please add a test that creates an object from a piece of data with known ID and patches a link prop on of of the links. I think what will happen now is: save() would just write the new linkprop value nulling all other linkprops on that link. Perhaps in cases like this save() should fetch linkprops before updating... We need to ponder on this.

PR #783, opened 2025-07-12
models: Fixes for link properties

elprans:
Fixes for handling of link properties in reflection and the query builder:

- **models: Implement `__linkprops__` class/instance duality**
- **models: Make linkprops actually work in qb expressions**
- **models: Move around model definitions in _qbmodel._abstract**
- **qb: Fix auto-injection of link properties into select shapes**

Fixes: #779


gel/_internal/_qbmodel/_pydantic/_models.py
```diff
@@ -1197,15 +1177,15 @@ class _MergedModelBase(GelModel, metaclass=_MergedModelMeta):
...
-    Generic[_MT_co],
+    _abstract.AbstractGelProxyModel[_MT_co, GelLinkModel],
     __gel_root_class__=True,
 ):
     __gel_has_id_field__ = False
 
     if TYPE_CHECKING:
         __gel_proxy_merged_model_cache__: ClassVar[type[pydantic.BaseModel]]
 
-    __slots__ = ("__linkprops__", "_p__obj__")
```
1st1, line 1208: Interesting. I guess this can fix serialization with exclude/include filters? why did you do this?

gel/_internal/_qbmodel/_pydantic/_models.py
```diff
@@ -1240,14 +1219,9 @@ def __init__(
         wrapped = self.__proxy_of__(id, **kwargs)
         ll_setattr(self, "_p__obj__", wrapped)
 
-        ll_setattr(
```
1st1, line 1243: why this change?

elprans: This was moved into the descriptor's `__get__`, makes more sense to do it there.

gel/_internal/_qbmodel/_pydantic/_models.py
elprans: Slots are incompatible with descriptors.

gel/_internal/_qbmodel/_pydantic/_models.py
1st1: add a comment to explain where you set the field

1st1 (approved):

Issue #794, opened 2025-07-15
gel generate fails on repeated runs


vpetrovykh:
**Describe the bug**
When using `gel generate py/models` after a schema update, the tools fails with an error.

**Reproduction**
1. Initialize a new project `gel init`
2. Make a simple schema `module default { type Foo }`.
3. run `gel generate py/models`
4. Change the schema  to `module default { type Foo {val: str} }`.
5. run `gel generate py/models`

```
$ gel init
No `gel.toml` (or `edgedb.toml`) found in `/home/victor/dev/magicstack/xxx` or above
Initializing new project...
Checking Gel versions...
┌─────────────────────┬──────────────────────────────────────────┐
│ Project directory   │ /home/victor/dev/magicstack/xxx          │
│ Project config      │ /home/victor/dev/magicstack/xxx/gel.toml │
│ Schema dir (empty)  │ /home/victor/dev/magicstack/xxx/dbschema │
│ Installation method │ portable package                         │
│ Version             │ 6.9+77bc1f3                              │
│ Instance name       │ xxx                                      │
│ Branch              │ main                                     │
└─────────────────────┴──────────────────────────────────────────┘
Version 6.9+77bc1f3 is already downloaded
Initializing Gel instance 'xxx'...
[edgedb] CRITICAL 281149 - 2025-07-15T17:49:46.849 postgres: the database system is starting up
[systemctl] Created symlink '/home/victor/.config/systemd/user/default.target.wants/edgedb-server@xxx.socket' → '/home/victor/.config/systemd/user/edgedb-server@xxx.socket'.
[systemctl] Created symlink '/home/victor/.config/systemd/user/default.target.wants/edgedb-server@xxx.service' → '/home/victor/.config/systemd/user/edgedb-server@xxx.service'.
Applying migrations...
Everything is up to date. Revision initial
Project initialized.
To connect to xxx, run `gel`
$ echo "module default { type Foo }" > dbschema/default.gel 
$ gel migration create --non-interactive
$ gel migration apply
Automatic backup is enabled, see the full backup list with "gel instance list-backups -I xxx".
Read more at https://geldata.com/p/localdev
Running automatic backup of local instance xxx...
Created full backup:
24.45 MiB /home/victor/.local/share/edgedb/data/xxx.backups/01981011-67cf-7070-ab16-cad7be488745
[...]

PR #798, opened 2025-07-16
models: Fix deserialization of introspected schema from cache

elprans:
Fix bugs in `_dataclass_extras._coerce_to_dataclass` and restore the
`AnyType` union as it is required for automatic discrimination of
dicts by a key.

Fixes: #794


1st1 (approved):

vpetrovykh (approved):
I tested it on the cache restoring case that was failing prior to this fix. It works now.

Issue #780, opened 2025-07-11
Implement `__bool__` that errors out on models


1st1:
* `Model.__bool__()` should error out suggesting to use `or_()`, `and_()`, et al. It should also say that the reason for this is that bool operators cannot be overloaded in Python.
* `Model.__contains__().__bool__()` should raise an error with a clear message suggesting to wrap expression in `not_()`
* We should implement code object introspection and ban use of `is` inside lambdas. This might be inconvenient for some users in some cases (using `is` on objects that are not models), but this footgun is just too big to ignore IMO.

PR #800, opened 2025-07-17
models: Ban `<expr> is [not] None`

elprans:
Similarly to prohibiting `<expr> == None`, raise on `<expr> is None`
is used.  Unlike boolean ops, there is no idiomatic way of intercepting
identity checks, so we have to resort to AST rewrites, which limits the
scope of where the protection applies to lambdas in query builder
expreesssions.

Follow up to PR #799 (issue #780).

### DIFF
--- a/gel/_internal/_is_overload.py
+++ b/gel/_internal/_is_overload.py
@@ -177,7 +177,7 @@ class _IsTransformer(ast.NodeTransformer):
         """
         return self._num_transforms
 
-    def visit_Compare(self, node: ast.Compare) -> ast.AST:  # noqa: N802
+    def visit_Compare(self, node: ast.Compare) -> ast.AST:
         """Visit a `Compare` node and transform `is` and `is not`
         operations.
         """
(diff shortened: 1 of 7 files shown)


gel/_internal/_is_overload.py
```diff
@@ -0,0 +1,209 @@
...
+    Args:
+        func: The function to be transformed.
+
+    Returns:
+        The transformed function.
+    """
+    if not callable(func):
+        return func
+
+    if getattr(func, "__gel_is_overloaded__", False):
```
1st1, line 55: can we find a way to cache by code object? file location + hash of `__code__` object bytes?

1st1: Image

1st1: basically right now you'd be rewriting that ast every time, we want to do that once and then reuse the rewritten code object (need to be careful with closure vars)

1st1 (approved):
I've some concerns about the efficiency of this. Currently it would be doing expensive code rewrite every time the query is built. I think it should be possible to maintain a cached mapping of `__code__` to rewritten `__code__` to avoid rewrites. The current "decorator-like" checkking wouldn't work for inline-lambda (or locally defined `def`) functions. See my comment.

PR #808, opened 2025-07-25
gel-generate-py models --output to relative path

### LINKED ISSUE #806: `gel-generate-py` should run from the current working directory
When you invoke the generator, it currently traverses back to "the root of the project" determined by where the `gel.toml` file lives. So when you pass an option like `--output`, the directory is relative to the root rather than the current working directory.

Given this common monorepo layout:

```
.
├── backend
│   ├── app
│   ├── pyproject.toml
│   └── tests
├── dbschema
│   └── default.gel
├── frontend
│   ├── package.json
│   ├── src
│   └── tests
└── gel.toml
```

If you run:
```
cd backend
gel-generate-py models --output ./app/models
```

It will actually generate into `/app/models` instead of `/backend/app/models`, which is unexpected from the standpoint of the caller.

fantix:
Fixes #806

### DIFF
--- a/gel/_internal/_codegen/_models/_pydantic.py
+++ b/gel/_internal/_codegen/_models/_pydantic.py
@@ -181,12 +181,14 @@ class PydanticModelsGenerator(AbstractCodeGenerator):
             logger.exception("could not connect to Gel instance")
             self.abort(61)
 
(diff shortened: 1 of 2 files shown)


scotttrinh (approved):
🙏

PR #813, opened 2025-07-29
Add query_graphql_json methods

msullivan:
There is not a non-JSON method, since making that work will be tricky
for annoying reasons, but I want to have it eventually.


msullivan:
Tests are failing because gel nightly is broken so doesn't support the feature yet

gel/abstract.py
```diff
@@ -597,6 +597,31 @@ def query_sql(
             )
         )
 
+    def query_graphql_json(
+        self,
+        query: str | Queryable[_T_ql],
```
elprans, line 602: The `Queryable` protocol is for EdgeQL, so these should only accept `str`.

PR #815, opened 2025-07-29
Apply generate config from env

fantix:
Parse and apply generate config from environment variables, see also geldata/gel-cli#1703.

This PR adds the infrastructure to parse generate config envs and report errors, as well as 2 configs for the `models` generator:

* `no_cache`: boolean, the same as `--no-cache`
* `output`: string, the same as `--output`

Parse error looks like:

image

### DIFF
--- a/gel/_internal/_codegen/_generator.py
+++ b/gel/_internal/_codegen/_generator.py
@@ -82,67 +82,105 @@ class AbstractCodeGenerator:
             self._quiet = args.quiet
 
     def _apply_env_config(self) -> None:
-        config_str = os.getenv("GEL_GENERATE_CONFIG")
-        if not config_str:
+        """
+        Apply environment configuration.
+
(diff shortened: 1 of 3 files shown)


fantix:
(I should probably reuse Python stdlib traceback formatting)

gel/_internal/_codegen/_generator.py
```diff
@@ -66,11 +67,88 @@ def __init__(
...
+                self.print_msg(
+                    f"{C.WARNING}Skipping unknown environment config: "
+                    f"{key}{C.ENDC}"
+                )
+                continue
+            try:
+                m(value["value"])
+            except ValueError as e:
+                span = value.get("span")
+                src = os.getenv("GEL_TOML_CONTENT")
```
1st1, line 108: This one is coming from Rust, right? prefix with an underscore?

1st1 (changes requested):
Downgrading my review to "some changes are needed"

1st1 (approved):

PR #817, opened 2025-07-30
Fix save() for link props on multi links

1st1:
Link props can now be partially updated, so in the scenario they are partially fetched or come from a manually instantiated model things should work as expected.


vpetrovykh (approved):
I'd say this looks good as a baseline. Let's merge and then fix corner cases.

elprans (approved):

PR #840, opened 2025-08-06
Add a high-level internals outline / onboarding guide

1st1:


gel/_internal/arch.md
```diff
@@ -0,0 +1,207 @@
...
+
+Run `python tools/gen_models.py` to generate test models into your virtual environment's site-packages. This enables IDE support for test development.
+
+## Development Setup
+
+### Prerequisites
+
+1. gel development VM with gel server binary in PATH
+2. Install with: `pip install -e .`
+3. **Critical**: Edit `.pth` files if pip editable install is broken
```
elprans, line 125: This is not an `if`, it is definitely broken because type checkers cannot "see" through the mumbo-jumbo setuptools generate for editable wheels. 

```suggestion
3. **Critical**: Edit `.pth` files because editable wheel install is broken:

python -c 'import pathlib, gel; print(pathlib.Path(gel.__path__[0]).parent)' > $(python -c 'import site; print(site.getsitepackages()[0])')/gel.pth 
```

Issue #832, opened 2025-08-06
Don't depend on type ids in models


msullivan:
The type ids of object types in gel are not consistent between instances. We must not hardcode them into generated code as anything but a potential optimization.

PR #847, opened 2025-08-08
Use `__tname__` instead of `__tid__` for polymorphic result resolution

elprans:
The type ids of object types in gel are not consistent between
instances. We must not hardcode them into generated code.

Fixes: #832


1st1 (approved):
good stuff, please remove a line in protocol.pyx marked with XXX, see my comment

PR #852, opened 2025-08-08
Support std.and_ and std.or_ with 2+ arguments

### LINKED ISSUE #843: `std.and_` and `std.or_` should accept 2+ number of args (now it's just 2 which is inconvenient)

aljazerzen:
Closes #843


gel/_internal/_codegen/_models/_pydantic.py
```diff
@@ -5661,6 +5669,90 @@ def get_ptr_type(
...
+    def _write_std_and(self):
+        self.write(textwrap.dedent("""\
+            def and_(  # type: ignore [override, unused-ignore]
+                a: base_shapes.bool | builtins.bool | type[base_shapes.bool],
+                b: base_shapes.bool | builtins.bool | type[base_shapes.bool],
+                *args: Sequence[
+                    base_shapes.bool | builtins.bool | type[base_shapes.bool
+                ]],
+            ) -> type[base_shapes.bool]:
+                match a:
```
1st1, line 5681: is using match here necessary? You're hardcoding this implementation, why not just if?

elprans (changes requested):
That's not the way, emitting source like that would almost certainly break.  You should special-case [`_write_infix_op_func_node_ctor`](https://github.com/geldata/gel-python/blob/c155f0ad19d8a5b5f1148f3e957d6a54fee3df9a/gel/_internal/_codegen/_models/_pydantic.py#L2740) instead.

elprans:
Oh, and if you special-case the AST node constructor you should be able to do the variadic injection at introspection as well, this should be the most robust solution.

flake.nix
```diff
@@ -0,0 +1,44 @@
...
+    }:
+    flake-utils.lib.eachDefaultSystem (
+      system:
+      let
+        pkgs = nixpkgs.legacyPackages.${system};
+        gel_pkgs = edgedb.packages.${system};
+      in
+      {
+        devShells.default = pkgs.mkShell {
+          venvDir = "/home/aljaz/EdgeDB/edgedb-python/.venv";
```
msullivan, line 26: We can't check this in

msullivan (approved):

aljazerzen:
Builds are failing because of https://github.com/geldata/gel-python/pull/891

aljazerzen:
> Merging is blocked. 1 review requesting changes and 1 approving review by reviewers with write access.

@elprans can you review?

PR #854, opened 2025-08-08
fastapi: reload app after Gel config update

fantix:
Use [the config hook](https://github.com/geldata/gel-cli/pull/1726) to trigger a FastAPI app reload, so that the user could for example:

1. User updates `gel.local.toml` to modify the SMTP server port
2. `gel watch --sync` applies such config
3. Config hook runs to modify a signal file
4. `fastapi dev` is woken to reload the app

### DIFF
--- a/gel/_internal/_qbmodel/_abstract/__init__.py
+++ b/gel/_internal/_qbmodel/_abstract/__init__.py
@@ -22,6 +22,7 @@ from ._descriptors import (
     AnyLinkDescriptor,
     AnyPropertyDescriptor,
     ComputedMultiLinkDescriptor,
+    ComputedMultiLinkWithPropsDescriptor,
     ComputedMultiPropertyDescriptor,
     ComputedPropertyDescriptor,
     GelLinkModelDescriptor,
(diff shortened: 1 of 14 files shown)


gel/_internal/_integration/_fastapi/_cli/_lifespan.py
```diff
@@ -102,33 +102,47 @@ def __exit__(
...
+        output = app_path / "models"
+        path = pathlib.Path(tmpdir)
+        reloader = str(path / "reload_fastapi.py")
+        reload_code = f"""\
+            with open("{reloader}", "w+t") as f: print(file=f)
+        """.strip().replace('"', '\\"')
+        content = textwrap.dedent(f"""\
+            [hooks-extend]
+            schema.update.after="gel-generate-py models --output={output}"
+            config.update.after="{sys.executable} -c '{reload_code}'"
```
elprans, line 116: Which shell do these hooks run in?  On Windows?

1st1 (approved):

Issue #848, opened 2025-08-08
Save vs Sync


vpetrovykh:
# Save vs Sync

There are two fundamentally different ways of persisting ORM model changes:

1. **One-way save**: Push changes to the database with minimal feedback (only new object IDs). Useful when you don't need updated computed values, defaults, or triggers.

2. **Bidirectional sync**: Push changes to the database, then refetch objects to get current state including computed values, defaults, and trigger effects.

Currently, `save(refetch=False)` handles case 1, while `save(refetch=True)` handles case 2. However, using the same method name for these different operations creates poor developer experience and the refetch mechanism is poorly defined.

The proposal is to separate these cases into 2 client methods: `save()` and `sync()`.

## Save

The `save()` method is the simpler of the two. The current implementation of `save(refetch=False)` already does what we need. Now it will be the default (and only) behavior of `save()`.

## Sync

The `sync()` method has 2 stages: save and re-fetch. The save stage works pretty much identically to `save()`. We apply the changes to the database and we keep track of new IDs. However, in addition to new IDs, we also record all the IDs of the Gel objects that were updated. We will refer to all of the new and updated object IDs as *delta*.

The next step is to refetch data from Gel. We need to define which Python objects and fields to refetch.

### Which Python objects need re-fetching

Any object that is either directly passed to `sync()` or is reachable by links from these root objects will need to be re-fetched. The reasoning is that these are the same objects that get scanned for changes that need to be saved in the first place. But we will re-fetch objects even if there were no direct changes that needed to be saved for them because they might be affected by changes in other objects in the sync batch (e.g. because of backlink computeds).

### Which data needs re-fetching

Deciding which data to re-fetch for each object is tricky because we want to avoid accidental re-fetching of large amounts of data that is not used.

The rules are slightly different for new and existing objects.

**New objects:**
- Refetch all properties (single, multi, regular, computed)
- Refetch all single links
- Multi-links follow existing object strategy

**Existing objects:**
- Skip "unset" fields (never fetched originally)
- Refetch previously fetched properties and single links
- Reconcile multi-links with delta (see strategy below)

### Multi-link Refetch Strategy

Multi-links aren't refetched entirely to avoid performance issues. Instead, we reconcile existing data with the *delta* (new and updated object IDs) by using a filter.

The refetch filter includes:
- All existing link target IDs from the Python field
- All IDs from the *delta*

This captures both additions and removals to the multi-link.

[...]

PR #855, opened 2025-08-08
Implement client.sync()

1st1:
see issue #848

### DIFF
--- a/tests/test_model_generator.py
+++ b/tests/test_model_generator.py
@@ -3933,18 +3933,24 @@ class TestModelGenerator(tb.ModelTestCase):
         )
         self.client.sync(gr)
 
-        # After saving the group, Alice's groups backlink should be unset
-        # because Alice wasn't explicitly fetched with groups after the change
-        with self.assertRaisesRegex(AttributeError, "'groups' is not set"):
-            assert alice.groups is not None  # access field
+        a = [u for u in gr.users if u.id == alice.id][0]
+        self.assertEqual(a.groups, alice.groups)
 
-        with self.assertRaisesRegex(AttributeError, "'groups' is not set"):
-            a = [u for u in gr.users if u.id == alice.id][0]
-            assert a.groups is not None  # access field
+        fresh_alice_groups = set(
+            self.client.query(
+                """
+                    with w := (select User{groups} filter .id=<uuid>$0)
+                    select w.groups.id
+                """,
+                alice.id,
+            )
+        )
 
-        # But Alice herself should still be accessible
+        self.assertEqual({g.id for g in a.groups}, fresh_alice_groups)
+
         self.assertEqual(alice.name, "Alice")
 
+    @tb.xfail
     def test_modelgen_save_reload_links_06(self):
         from models import default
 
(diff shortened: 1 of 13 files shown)


tests/test_model_generator.py
```diff
@@ -3722,13 +3772,277 @@ def test_modelgen_save_reload_links_03(self):
...
+
+        with self.assertRaisesRegex(
+            AttributeError, "'comp_opt_friend' is not set"
+        ):
+            assert tsl2.comp_opt_friend is not None  # access field
+
+        with self.assertRaisesRegex(
+            AttributeError, "'comp_opt_wprop_friend' is not set"
+        ):
+            assert tsl2.comp_opt_wprop_friend is not None  # access field
```
vpetrovykh, line 3910: This was the result of a `sync(tsl2)`. The object `tsl2` was fetched from the database with computed links being included (so none of them are "unset"). This means that after the sync, we expect the computed links to still be present because they were explicitly fetched originally. So they should have been re-fetched.

tests/test_model_generator.py
```diff
@@ -3722,13 +3772,277 @@ def test_modelgen_save_reload_links_03(self):
...
+        self.client.sync(gr)
+
+        # After saving the group, Alice's groups backlink should be unset
+        # because Alice wasn't explicitly fetched with groups after the change
+        with self.assertRaisesRegex(AttributeError, "'groups' is not set"):
+            assert alice.groups is not None  # access field
+
+        with self.assertRaisesRegex(AttributeError, "'groups' is not set"):
+            a = [u for u in gr.users if u.id == alice.id][0]
+            assert a.groups is not None  # access field
```
vpetrovykh, line 3943: Similar to the previous case, `alice` has been explicitly fetched with the backlink groups. The Python object `alice` has been literally added to the new group's multi link `users`. So it is reachable by the sync from the root `gr`. Therefore the expectation is that the backlink of `alice` must be re-fetched because it was not "unset".

tests/test_model_generator.py
```diff
@@ -3722,13 +3772,277 @@ def test_modelgen_save_reload_links_03(self):
...
+        red.users.remove(alice)
+        self.client.sync(red)
+
+        # # After saving, Alice's groups backlink should be unset
+        # with self.assertRaisesRegex(AttributeError, "'groups' is not set"):
+        #     assert alice.groups is not None  # access field
+
+        # # But we should still have some valid data
+        # self.assertEqual(alice.name, "Alice")
+        # self.assertEqual({u.id for u in red.users}, orig_ids - {alice.id})
```
vpetrovykh, line 3985: This one is a little tricky because `alice` was definitely explicitly fetched with the backlink `groups` (not "unset"). Since it was removed from `red` it is not technically reachable from `red` anymore at the point of sync. However, I think we agreed that "reachable" should include any object being removed (since we know what they are in order to remove them). Therefore the expectation should be for the backlink of `alice` to be re-fetched.

tests/test_model_generator.py
1st1: ok I fixed this one

tests/test_model_generator.py
1st1: This one I *can* fix

1st1: ok pushed a fix it's working now

tests/test_model_generator.py
1st1: @msullivan

tests/test_model_generator.py
1st1: done, all working now

PR #858, opened 2025-08-09
Update distinct list tests.

### LINKED ISSUE #841: every test in test_distinct_list.py is npw xfailed -- we need good tests for `_link_set.py` and `_tracked_list.py`

dnwpark:
related #841

Updated dlist tests and separated them into new test files:
- `test_link_set_abstract.py` for `AbstractLinkSet` and `AbstractMutableLinkSet`
- `test_link_set.py` for concrete instantiations of `LinkSet`
- `test_link_with_props_set.py` for concrete instantiations of `LinkWithPropsSet`
- `test_link_set_models.py` for link sets generated from models

Some of the new tests are marked xfail:
- Tests for querying model generated link sets fail for three reasons:
    - The link is not fetched despite being marked `True` or a splat is used
    - An empty collection is produced
    - An error like: "expected ForwardRef to be a path alias"
- Tests for modifying model generated link sets using  `+=` and `-=` fail because mypy seems to be checking `+` and `-` instead

Additionally, all tests of the form `test_link_set_model_fresh_multi_link_with_props_` fail with `AttributeError: 'ModelFieldDescriptor' object has no attribute 'link'`.


tests/test_link_set.py
```diff
@@ -0,0 +1,245 @@
...
+    AbstractGelModel,
+)
+from gel._internal._qbmodel._abstract._link_set import (
+    LinkSet,
+)
+from gel._internal._tracked_list import Mode
+
+
+# Some test models
+class BoxedInt(AbstractGelModel):
```
msullivan, line 14: What do we need boxes for here? Because the containers want AbstractGelModels?

dnwpark: Yeah basically

PR #863, opened 2025-08-21
Add QueryRefetchArgs.

dnwpark:
A simple refactor for _save.


gel/_internal/_save.py
```diff
@@ -1156,11 +1156,21 @@ def feed_db_data(self, obj_data: Iterable[Any]) -> None:
                 pass
 
 
+@dataclasses.dataclass(kw_only=True, frozen=True)
+class QueryRefetchArgs:
+    spec: list[tuple[
+        uuid.UUID,
+        list[tuple[bool, list[uuid.UUID]]],
```
1st1, line 1163: Separately would appreciate you creating a `TypeAliasType` for spec with a comment on its structure.

PR #869, opened 2025-08-22
Refetch computed for newly inserted objects

dnwpark:
Given a schema
```
type A {
   total_val := sum(.<a[is B].val);
};
type B {
    a: A;
    val: int64;
};
```

If a user does:
```
a = default.A()
b = default.B(a=a, val=7)

client.sync(a, b)
```

The order of inserts will be `a`, then `b` and so the value of `a.total_val` fetched during the insert will be 0.

This PR adds a refetch query for the computed properties of newly inserted objects.

Tests are based on https://github.com/geldata/gel-python/issues/848

### DIFF
--- a/gel/_internal/_save.py
+++ b/gel/_internal/_save.py
@@ -643,18 +643,12 @@ def _add_refetch_shape(
                 ref_shape.fields[field_name] = ptr_info
 
     else:
-        # New objects only need computeds refetched
-        for field_name in obj.__pydantic_computed_fields__:
-            if field_name in ref_shape.fields:
+        # New objects only need computed properties refetched
+        for ptr_name, ptr_info in tp_pointers.items():
+            if not ptr_info.computed or ptr_info.kind != PointerKind.Property:
                 continue
 
-            try:
-                ptr_info = tp_pointers[field_name]
-            except KeyError:
-                # ad-hoc computed
-                pass
-            else:
-                ref_shape.fields[field_name] = ptr_info
+            ref_shape.fields[ptr_name] = ptr_info
 
 
 def push_refetch_new(
--- a/gel/_internal/_save.py
+++ b/gel/_internal/_save.py
@@ -625,25 +625,21 @@ def _add_refetch_shape(
     tp_pointers = tp_obj.__gel_reflection__.pointers
     ref_shape = refetch_ops[tp_obj]
     ref_shape.models.append(obj)
-    if not obj.__gel_new__:
-        # Existing objects should refetch anything that was previously set
-        for field_name in obj.__pydantic_fields_set__:
-            if (
-                field_name in ref_shape.fields
-                or field_name in _STOP_FIELDS_NO_ID
-            ):
-                continue
+    # Existing objects should refetch anything that was previously set
+    for field_name in obj.__pydantic_fields_set__:
+        if field_name in ref_shape.fields or field_name in _STOP_FIELDS_NO_ID:
+            continue
 
-            try:
-                ptr_info = tp_pointers[field_name]
-            except KeyError:
-                # ad-hoc computed
-                pass
-            else:
-                ref_shape.fields[field_name] = ptr_info
+        try:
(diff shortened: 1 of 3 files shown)


gel/_internal/_save.py
```diff
@@ -595,27 +595,60 @@ def model_attr(
...
+                ptr_info = tp_pointers[field_name]
+            except KeyError:
+                # ad-hoc computed
+                pass
+            else:
+                ref_shape.fields[field_name] = ptr_info
+
+    else:
+        # New objects only need computeds refetched
+        for field_name in obj.__pydantic_computed_fields__:
```
1st1, line 625: I'd use pointer introspection here, not `__pydantic_computed_fields__`

gel/_internal/_save.py
```diff
@@ -1399,7 +1447,7 @@ def _compile_refetch(
                 collections.defaultdict(dict)
             )
             for obj in shape.models:
-                link_ids = obj_links_all[obj.id]
+                link_ids = obj_links_all[self._get_id(obj)]
```
1st1, line 1450: good catch

gel/_internal/_save.py
```diff
@@ -1303,6 +1350,7 @@ class SaveExecutor:
 
     def __post_init__(self) -> None:
         self.new_object_ids = {}
+        self.new_objects = {}
```
1st1, line 1353: Ouch. You'll need to rebase this onto the latest master -- do it now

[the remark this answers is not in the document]
1st1: please rebase onto the fresh `master`

gel/_internal/_save.py
```diff
@@ -1625,6 +1675,18 @@ def _traverse(obj: GelModel) -> None:
                         # see the above comment.
                         multi_prop_commit_recursive(linked)
 
+            if new_obj_id := self.new_object_ids.get(id(obj)):
```
1st1, line 1678: yeah, rebase is needed. Basically with the latest `master` you'd just `self.new_object_ids.get(obj)`

gel/_internal/_save.py
```diff
@@ -595,27 +595,60 @@ def model_attr(
...
+            try:
+                ptr_info = tp_pointers[field_name]
+            except KeyError:
+                # ad-hoc computed
+                pass
+            else:
+                ref_shape.fields[field_name] = ptr_info
+
+    else:
+        # New objects only need computeds refetched
```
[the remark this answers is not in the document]
msullivan: Not by a trigger, but by a rewrite certainly. We probably do want to refetch all of them?

(Which design do you mean here?)

dnwpark:
This now includes fixes for #884

gel/_internal/_save.py
```diff
@@ -1671,6 +1710,29 @@ def _traverse(obj: GelModel) -> None:
...
+                and (new_obj_id := self.new_object_ids.get(obj))
+            ):
+                new_obj = self.new_objects.get(new_obj_id)
+
+                # Update new objects with refetched data
+                for prop in get_pointers(type(obj)):
+                    if (
+                        # prop not refetched
+                        prop.name not in new_obj.__dict__
+                        # TODO: Refetching links for new objects
```
msullivan, line 1725: ?

gel/_internal/_save.py
```diff
@@ -1671,6 +1710,29 @@ def _traverse(obj: GelModel) -> None:
                         # see the above comment.
                         multi_prop_commit_recursive(linked)
 
+            if (
+                self.refetch
+                and obj in self.new_object_ids
+                and (new_obj_id := self.new_object_ids.get(obj))
+            ):
+                new_obj = self.new_objects.get(new_obj_id)
```
msullivan, line 1718: What's going on here? Why does new_objects need a different path for updating the objects with refetched data than the old objects?

msullivan (approved):

PR #870, opened 2025-08-22
fastapi auth: reuse verifier to continue the flow

nsidnev:
If email confirmation is required to complete the sign-up , the verifier will be overwritten after being redirected to the sign-in page, resulting in an error when exchanging tokens when following the link from the email


fantix (approved):
Awesome! That's almost identical to my local change

PR #871, opened 2025-08-25
Make model generation more (ideally fully?) deterministic

msullivan:
The generation was not very deterministic, and the reproducibility test
wasn't catching it.  Dropping and recreating the database between cycles
exposed a lot of issues (including still embedding ids).

 * Don't embed id in schema object values. Put a dummy in instead.
  (Will this be a problem?)
* Sort groups of potentially overlapping functions by their
  CallableSignatures
* Inside overlapping groups, sort by edgeql_signature in addition
  to their "generality_key"
* Don't use base64 of ids to generate Tuple names. Instead use
  a base64 of name. This produces kind of horrific names; at some
  point we should try to safely produce something based on the
  actual name.

### DIFF
--- a/gel/_internal/_codegen/_models/_pydantic.py
+++ b/gel/_internal/_codegen/_models/_pydantic.py
@@ -1023,7 +1023,8 @@ class BaseGeneratedModule:
                 )
                 self.write(f"return {objtype_mod}.ObjectType(")
                 with self.indented():
-                    self.write(f"id={uuid_}(int={objtype.uuid.int}),")
+                    self.write(f"# Dummy id, for stability/portability")
+                    self.write(f"id={uuid_}(int=0),")
                     self.write(f"name={objtype.name!r},")
                     self.write(f"builtin={objtype.builtin!r},")
                     self.write(f"internal={objtype.internal!r},")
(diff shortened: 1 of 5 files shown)


gel/_internal/_codegen/_models/_pydantic.py
```diff
@@ -1277,7 +1278,11 @@ def get_tuple_name(
         t: reflection.NamedTupleType,
     ) -> str:
         names = [elem.name.capitalize() for elem in t.tuple_elements]
-        digest = base64.b64encode(t.uuid.bytes[:4], altchars=b"__").decode()
+        # Base64ing the name is pretty horrific but it avoids needing
+        # to escape it into Python safely.
```
1st1, line 1282: Hm, I don't quite understand this -- the name would be a string, right? why not just `repr()` it?

msullivan: It won't be a valid python identifier

msullivan: It will be something like `tuple<a: std::int64, b: array<tuple<std::str, std::bool>>>` or whatever

msullivan: Sure

elprans (approved):

PR #877, opened 2025-08-27
Fix clear on multi prop not removing old values after sync.

dnwpark:
related #848

Given a schema:
```
type A {
    multi val: int64;
};
```

Running the script:
```
my_a = default.A(val=[1, 2, 3])
self.client.save(my_a)

my_a.val = [4, 5, 6]
self.client.sync(my_a)
```

Would result in `my_a.val` being `[1, 2, 3, 4, 5, 6]` instead of `[4, 5, 6]`.


gel/_internal/_tracked_list.py
```diff
@@ -282,9 +282,7 @@ def _ensure_snapshot(self) -> None:
             self._removed_items = []
 
     def __gel_replace_with_empty__(self) -> None:
-        self._items.clear()
-        self._added_items = None
-        self._removed_items = None
+        self.clear()
```
msullivan, line 285: Why is there a special `__gel_replace_with_empty__` method anyway?
Isn't that just `clear`?

msullivan (approved):

PR #894, opened 2025-09-02
Allow running tests with geltest

elprans:
- **models: Add the ability to generate std separately and then import it**
- **Formalize server instance management a-la it is done in Gel server itself**
- **Make `IsolationLevel` a proper enum**
- **Allow running tests with geltest**


gel/_internal/_cache.py
```diff
@@ -8,56 +8,120 @@
...
+def load_text(
+    key: str,
+    *,
+    cache_dir: pathlib.Path | None = None,
+    extra_key: str | None = None,
+) -> str | None:
+    if cache_dir is None:
+        cache_dir = _default_cache_dir
+    if extra_key is not None:
+        _verify_cache(key, cache_dir=cache_dir, extra_key=extra_key)
```
msullivan, line 90: I don't think this does anything. `_verify_cache` returns whether the cache is legit but we are ignoring that.
Same with `load_bytes` and `load_json`

gel/_internal/_server.py
```diff
@@ -0,0 +1,1186 @@
+# SPDX-PackageName: gel-python
```
msullivan, line 1: If this is *only* needed for tests, then we should put it in a `_testbase` directory I think

fantix (approved):

PR #908, opened 2025-09-05
Update gen_models for changes to testbase.

dnwpark:


tests/test_link_set_models.py
```diff
@@ -2770,7 +2770,7 @@ def test_link_set_model_fresh_multi_link_with_props_update_01(self):
         )
         self.assertEqual(
             {f: f.__linkprops__.opinion for f in harry.friends},
-            {hermione: "smart", ron: "reliable"},
```
msullivan, line 2773: Why did this change here?

msullivan (approved):
Do we even need gen_models anymore, though?

tests/test_link_set_models.py
dnwpark: The behaviour changed in  #903 to assume that `update` and `+=` wraps links with a new proxy model instead of "merging" the link proxy. Not sure why the test didn't fail there though...

PR #914, opened 2025-09-16
Add readme for _save.

dnwpark:


msullivan (approved):
Great!

Maybe also say something about some of the horrendous hacks for representing optional fields in tuples? (putting them in arrays, etc)

PR #918, opened 2025-09-18
Embed __tname__ in models, support using it in model_validate

### LINKED ISSUE #755: support model_validate
We currently don't support `model_validate` but we should, it's one of the most used Pydantic APIs. For the most part the following does the trick:

```diff
diff --git a/gel/_internal/_qbmodel/_pydantic/_fields.py b/gel/_internal/_qbmodel/_pydantic/_fields.py
index 4656af7..7ea9cd6 100644
--- a/gel/_internal/_qbmodel/_pydantic/_fields.py
+++ b/gel/_internal/_qbmodel/_pydantic/_fields.py
@@ -70,6 +70,15 @@ class _MultiPointer(_abstract.PointerDescriptor[_T_co, _BT_co]):
     ) -> Any:
         raise NotImplementedError

+    @classmethod
+    def _get_final_type(
+        cls,
+        generic_args: tuple[type[Any], type[Any]],
+    ) -> type[Any]:
+        raise NotImplementedError(
+            f"{cls.__name__} is missing _get_final_type() implementation"
+        )
+
     @classmethod
     def __get_pydantic_core_schema__(
         cls,
@@ -78,11 +87,29 @@ class _MultiPointer(_abstract.PointerDescriptor[_T_co, _BT_co]):
     ) -> pydantic_core.CoreSchema:
         if _typing_inspect.is_generic_alias(source_type):
             args = typing.get_args(source_type)
-            return core_schema.no_info_plain_validator_function(
-                functools.partial(cls._validate, generic_args=args),
-                serialization=core_schema.plain_serializer_function_ser_schema(
-                    list,
-                ),
+            final_type = cls._get_final_type(args)
+
+            inner_schema = handler.generate_schema(args[0])
+            return core_schema.union_schema(
+                [
+                    core_schema.is_instance_schema(
+                        final_type,
+                        cls_repr=final_type.__name__,
+                    ),
+                    core_schema.chain_schema(
+                        [
+                            core_schema.list_schema(inner_schema),
+                            core_schema.no_info_plain_validator_function(
+                                functools.partial(
+                                    cls._validate, generic_args=args
+                                ),
+                                serialization=core_schema.plain_serializer_function_ser_schema(
+                                    list,
+                                ),
+                            ),
+                        ]
+                    ),
+                ]
             )
         else:
             return handler.generate_schema(source_type)
```

however! Patterns that involve inheritance don't work out of the box:

```python
a = models.Account.model_validate(
  {"username": "aaa", "watchlist": [{"title": "bb"}]}
)
print(a)
# id=<UUID: UNSET> username='aaa' watchlist=[Content(title='bb')]
```

so basically the base type got picked up and used.

This would be exact same problem when vanilla pydantic is used naively:

```python
class Content(pydantic.BaseModel):
    title: str

class Movie(Content):
    pass

class TVShow(Content):
    pass

class Account(pydantic.BaseModel):
    username: str
    watchlist: list[Content]

a = models.Account.model_validate(
  {"username": "aaa", "watchlist": [{"title": "bb"}]}
)
print(a)
# username='aaa' watchlist=[Content(title='bb')]
```

However, pydantic has mechanisms to specify how unions should be discriminated: https://docs.pydantic.dev/latest/concepts/unions/

```python
class Content(pydantic.BaseModel):
    title: str
    kind: typing.Literal["movie", "tvshow"]


class Movie(Content):
    kind: typing.Literal["movie"]


class TVShow(Content):
    kind: typing.Literal["tvshow"]


class Account(pydantic.BaseModel):
    username: str
    watchlist: list[
        typing.Annotated[
            Movie | TVShow,
            pydantic.Field(
                discriminator="kind",
            ),
        ]
    ]

a = Account.model_validate(
    {"username": "aaa", "watchlist": [{"kind": "movie", "title": "bb"}]}
)

print(a)
# username='aaa' watchlist=[Movie(title='bb', kind='movie')]
```

We need to discuss what exactly do we do here: 
* always require `__type__` passed explicitly? 
* add another schema-level annotation to tag a field used to discriminate?
* add a schema-level union tag to make it more formal and validated than an annotation?
* do both? other options?

msullivan:
Gel supports inheritance much more than pydantic does, and so when we
have a link to a parent type, we need to be able to determine which of
its potential children type to create.

We do this by adding a `tname__` field to every generated class,
renamed to `__tname__` via the alias system, and update
`__gel_validate__` (which currently just directly calls
`model_validate`) to discriminate on it. The hard part here is
handling the case where there are linkprops, which are represented by
per-link subclasses of ProxyModel.  To deal with this, we dynamically
generate subclasses of the link class.

This mechanism also allows us to return correct types from the client
for links with link properties that point to subtypes, which we
previously weren't.

The other potential option was to use pydantic's discriminated unions,
either by directly generating them in the source or dynamically
mocking them up when creating models.  Getting this to play nicely
with everything we do seemed pretty nasty. (The linkprop situation in
particular scares me more.) The big potential advantage of doing it
that way, instead of having our own custom callbacks, is that it might
allow us to not lose track of the parameters of validation (like
strict mode and whether we are in JSON mode).

But it turns out we were already losing track of that due to
various ways we were interposing ourselves in validation, so
maybe it's fine. We can maybe fix this with contextvars later.
(Though the json situation is rough.)

Fixes #755.


msullivan:
Don't review individual commits, it is useful, but I left them in case I wanted to bisect something before I merge.

gel/_internal/_qbmodel/_abstract/_base.py
```diff
@@ -162,8 +165,13 @@ def __new__(
             super().__new__(mcls, name, bases, namespace, **kwargs),
         )
         reflection = cls.__gel_reflection__
-        if (tname := getattr(reflection, "name", None)) is not None:
-            mcls.__gel_class_registry__[tname] = cls
+        if (
+            (tname := getattr(reflection, "name", None)) is not None
+            and not issubclass(cls, AbstractGelLinkModel)
+            and not hasattr(cls, '__linkprops__')
+            and not hasattr(cls, '__gel_merged_model__')
```
elprans, line 172: It is probably more correct to check for `__gel_root_class__` which we should uplift from `_pydantic.GelModelMeta`.

gel/_internal/_qbmodel/_pydantic/_utils.py
```diff
@@ -201,6 +208,14 @@ def massage_model_dump_kwargs(
     except KeyError:
         exclude = kwargs["exclude"] = set()
 
+    # HACK
```
elprans, line 211: Please elaborate

elprans (approved):

gel/_internal/_qbmodel/_abstract/_base.py
msullivan: I changed this mechanism to check for the presence of `tname__` in namespace, which picks up all of these cases more cleanly and covers two cases I had missed `__build_custom_serializer` and user-defined subtypes

gel/_internal/_qbmodel/_pydantic/_models.py
```diff
@@ -1373,6 +1421,12 @@ class ProxyModel(
     if TYPE_CHECKING:
         __gel_proxy_merged_model_cache__: ClassVar[type[_MergedModelBase]]
 
+    # A cache for dynamically generated proxies where the __proxy_of__
+    # class is a *subtype*. Generated by _get_subtype_proxy.
+    __gel_subtype_proxy_cache__: ClassVar[
```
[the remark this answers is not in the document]
msullivan: I think the keys typically won't ever die

PR #926, opened 2025-09-24
Move tname__ to base_shapes to reduce noise in interface classes.

dnwpark:
followup to #918

`tname__` was added to the user facing types. For example, in `models.orm.default`:
```py
class User(base_shapes.User, Named):
    """type default::User"""

    tname__: Literal["default::User"] = Field("default::User", alias="__tname__")
    groups: ComputedMultiLink[UserGroup]
```

This PR would moves `tname__` to `base_shapes.User`.

### DIFF
--- a/gel/_internal/_qbmodel/_abstract/_base.py
+++ b/gel/_internal/_qbmodel/_abstract/_base.py
@@ -167,9 +167,28 @@ class AbstractGelModelMeta(GelTypeMeta):
         reflection = cls.__gel_reflection__
         if (
             # The class registry only tracks the canonical base instances,
-            # which are the ones that directly declare 'tname__'
+            # which are the ones with a base that directly declares 'tname__'
+            # but do not declare it themselves.
+            #
+            # Eg.
+            #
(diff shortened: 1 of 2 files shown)


gel/_internal/_qbmodel/_abstract/_base.py
```diff
@@ -169,7 +169,7 @@ def __new__(
             # The class registry only tracks the canonical base instances,
             # which are the ones that directly declare 'tname__'
             (tname := getattr(reflection, "name", None)) is not None
-            and 'tname__' in namespace
+            and any('tname__' in base.__dict__ for base in bases)
```
msullivan, line 172: Sigh I guess this is fine.

Also update the comment above this, which is now wrong.

Add that `tname__` isn't in `namespace` itself, then, I think?
Since the base shape also has base shape children, and we don't want those to get picked up.
(I don't know if it would be a bug for them to get picked up here, since they'd be correctly overwritten by their defined-later real children, but still.)

msullivan (approved):

PR #929, opened 2025-09-25
Apply splat to schema sets and path aliases without explicit select.

### LINKED ISSUE #838: Make the query builder behave consistently w.r.t. having or not having `.select()`
**Describe the bug**
Currently if you omit the `.select()` in QB you get different behavior depending on whether it's top-level or not.

```
User.select(
    groups=User.groups.filter(name="green")
)
```
Will be equivalent to:
```
select User {
  *,
  groups: {id} filter .name = 'green',
}
```

**Expected behavior**
We want this to be consistent. Omitting an explicit `.select()` should always be equivalent to doing `.select('*')`.
This:
```
User.filter(...)
User.select(
    groups=User.groups.filter(name="green")
)
```
should be equivalent to:
```
User.select("*").filter(...)
User.select(
    groups=User.groups.select("*").filter(name="green")
)
```

dnwpark:
close #838 

Ensures a splat is applied in the following cases:
```py
default.User
default.User.filter(...)
default.User.groups
default.User.groups.filter(...)
default.User.filter(...).groups
default.User.filter(...).groups.filter(...)
```

Also ensures that splats are applied if such a case is used within another object's select.
```py
default.Group.select(
    users=default.Post.author  # splat applied here
)
```

### DIFF
--- a/gel/_internal/_qbmodel/_abstract/_expressions.py
+++ b/gel/_internal/_qbmodel/_abstract/_expressions.py
@@ -192,19 +192,18 @@ def select(
             if isinstance(el_expr, (_qb.SchemaSet, _qb.Path)):
                 # If the expression is a schema set or path without an explicit
                 # select, apply a splat.
-                if isinstance(kwarg, type) and issubclass(
-                    kwarg, _qb.GelObjectTypeMetadata
-                ):
-                    el_expr = _qb.ShapeOp(
-                        iter_expr=el_expr,
-                        shape=_qb.get_object_type_splat(kwarg),
+                if (
+                    el_type := (
+                        kwarg
+                        if isinstance(kwarg, type)
+                        else kwarg.__gel_origin__
+                        if isinstance(kwarg, _qb.BaseAlias)
+                        else None
                     )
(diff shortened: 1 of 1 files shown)


msullivan (approved):

gel/_internal/_qbmodel/_abstract/_expressions.py
```diff
@@ -189,6 +189,24 @@ def select(
...
+                if isinstance(kwarg, type) and issubclass(
+                    kwarg, _qb.GelObjectTypeMetadata
+                ):
+                    el_expr = _qb.ShapeOp(
+                        iter_expr=el_expr,
+                        shape=_qb.get_object_type_splat(kwarg),
+                    )
+                elif isinstance(kwarg, _qb.BaseAlias) and issubclass(
+                    kwarg.__gel_origin__, _qb.GelObjectTypeMetadata
+                ):
```
msullivan, line 204: I would try to merge these cases, instead of duplicating the bodies?

msullivan (approved):

PR #930, opened 2025-09-25
Add xfail tests for links to std::Object not returning the correct type.

dnwpark:


msullivan (approved):
Mention in the xfail calls what is wrong

PR #944, opened 2025-10-02
Add cast function.

dnwpark:
Allows users to cast between types:
```py
assert client.get(std.str.cast(std.int64(1))) == "1"
assert client.get(std.array[std.str].cast(
    std.array[std.int64](
        [std.int64(1), std.int64(2), std.int64(3)]
    )
) == ["1", "2", "3"]
```


msullivan:
Is the idea here that we support generating casts between any two types, and then we let the server deal with figuring out which are real?

That's probably a reasonable first take.
I think eventually we might want to generate overloads based on reflecting the casts in the schema, but this is an expedient first step.

@elprans @1st1 thoughts?

msullivan (approved):

tests/test_qb.py
```diff
@@ -1548,3 +1650,24 @@ def test_qb_enum_edit_02(self):
...
+        from models.orm import default, std
+
+        result = self.client.get(std.str.cast(default.Color.Red))
+        self.assertEqual(result, "Red")
+
+    def test_qb_cast_scalar_03(self):
+        # scalar to enum
+        from models.orm import default, std
+
+        result = self.client.get(default.Color.cast(std.str("Red")))
```
elprans: This is an example of where I feel the explicit `cast` is inconsistent.  We are casting a Python literal string `"Red"` to `std.str` via a (Pythonic) call to a constructor, whereas the second cast is via a classmethod.

dnwpark:
Removing the extra python casts is a more general problem. Will be fixed elsewhere.

PR #948, opened 2025-10-10
Add type intersection expressions

dnwpark:
Given a schema:
```edgeql
type Inh_A { a: int64; };
type Inh_B { b: int64; };
type Inh_C { c: int64; };
type Inh_AB extending Inh_A, Inh_B { ab: int64; };
type Inh_AC extending Inh_A, Inh_C { ac: int64; };
type Inh_BC extending Inh_B, Inh_C { bc: int64; };
type Inh_ABC extending Inh_A, Inh_B, Inh_C { abc: int64; };
type Inh_AB_AC extending Inh_AB, Inh_AC { ab_ac: int64; };
```

Supports queries such as:
```py
client.query(
    default.Inh_A.when_type(default.Inh_B).is_(default.Inh_C)
)
client.query(
    default.Inh_AB.select(a=lambda x: x.is_(default.Inh_C).c)
)
```

### DIFF
--- a/gel/_internal/_qbmodel/_abstract/_methods.py
+++ b/gel/_internal/_qbmodel/_abstract/_methods.py
@@ -14,7 +14,7 @@ from typing import (
     TypeVar,
 )
 from typing_extensions import Self
-
+import weakref
 
 from gel._internal import _qb
 from gel._internal._schemapath import (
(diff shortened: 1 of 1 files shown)


gel/_internal/_qb/_expressions.py
```diff
@@ -300,6 +300,35 @@ def __edgeql_expr__(self, *, ctx: ScopeContext) -> str:
...
+
+    def __init__(
+        self,
+        *,
+        expr: ExprCompatible,
+        type_filter: TypeNameExpr,
+        type_: TypeNameExpr,
+    ) -> None:
+        op = _edgeql.Token.RANGBRACKET
+        object.__setattr__(self, "type_filter", type_filter)
```
msullivan: This is an issue for another PR (or not bothering), but all these __setattr__s and explicit `__init__`s in a dataclass are dodgy; are they all just so that `op` doesn't show up in the dataclass string like it would if it was declared with a default

gel/_internal/_qbmodel/_abstract/_methods.py
```diff
@@ -206,3 +236,176 @@ def __gel_assert_single__(
...
+        if (
+            (rhs_p_ref := rhs.get(p_name)) is not None
+            and (p_ref := process_common(lhs_p_ref, rhs_p_ref)) is not None
+        )
+    }
+
+    return result
+
+
+def create_intersection(
```
msullivan: Some comments about what is being done here exactly would be useful

gel/protocol/codecs/object.pyx
```diff
@@ -348,23 +348,30 @@ cdef class ObjectCodec(BaseNamedRecordCodec):
...
+        if proxy := getattr(return_type, '__proxy_of__', None):
             self.cached_return_type_proxy = return_type
             self.cached_return_type = proxy
             assert not hasattr(proxy, '__proxy_of__')
             lprops_type = return_type.__linkprops__
+            worklist = [self.cached_return_type]
+        elif expr_object_types is not None:
+            self.cached_return_type = return_type
+            self.cached_return_type_proxy = None
+            worklist = list(expr_object_types)
```
msullivan: Maybe name this something different than `worklist` here, and make a copy to populate the initial worklist?
(Instead of needing to recreate the `worklist` conditionally after the DFS loop)

msullivan (approved):
Need to find a new name probably though

